PR #10, opened 2019-02-17
Sync with latest grakn-core changes

vmax:
# Why is this PR needed?

So we are able to `bazel build` this repo

# What does the PR do?

- Upgrades commit hash of `@graknlabs_grakn_core`
- Adds missing dependencies for `@graknlabs_graql`

# Does it break backwards compatibility?

—

# List of future improvements not on this PR

—


haikalpribadi:
The tests seem to fail though, @vmax. I'll merge this first and we can investigate afterwards.

PR #26, opened 2019-03-05
Fix dependency on enum34

vmax:

Fix `client-python` on 3.4+

`enum34` is incompatible with Python 3.4+. Instead, we replace it with `enum-compat` which a no-op on incompatible versions


flyingsilverfin:
@vmax can we just get rid of this enum by any chance?

PR #35, opened 2019-03-25
Add benchmark as transitive dependency

flyingsilverfin:
Fix breaking change caused by the introduction of `benchmark` repository as a Bazel git_repository dependency.

One liner to include `graknlabs_benchmark()` dependency that allows the building of distribution `tar.gz` in CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,7 @@ jobs:
       - checkout
       - run: sudo apt-get update
       - run: sudo apt-get install python-pip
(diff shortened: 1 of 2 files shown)


haikalpribadi:
you need to add environment variables for the deployment test to pass (see below), but I'll merge this first as I know it will pass in `master`.
```
DEPLOYMENT_REPO_TYPE=test DEPLOYMENT_USERNAME=$REPO_GRAKN_USERNAME DEPLOYMENT_PASSWORD=$REPO_GRAKN_PASSWORD
```

PR #36, opened 2019-03-26
Release pipeline for client-python

vmax:

CI pipeline for releasing `client-python`

* Add `deploy_github` for posting GitHub draft release
* Wire CircleCI jobs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -157,7 +157,7 @@ workflows:
           filters:
             branches:
               only: client-python-release-branch
-      - release-approval:
+      - release-deployment-approval:
           type: approval
           requires:
             - release-github
@@ -166,7 +166,7 @@ workflows:
               only: client-python-release-branch
       - release-pypi:
           requires:
-            - release-approval
+            - release-deployment-approval
           filters:
             branches:
               only: client-python-release-branch
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,6 +99,14 @@ jobs:
       - checkout
       - run: bazel run @graknlabs_build_tools//ci:release-approval
 
+  release-github:
+    machine: true
+    working_directory: ~/client-python
+    steps:
+      - install-bazel-linux-rbe
+      - checkout
+      - run: bazel run //:deploy-github
+
   release-pypi:
     machine: true
     working_directory: ~/client-python
@@ -145,7 +153,20 @@ workflows:
 
   client-python-release:
     jobs:
+      - release-github:
+          filters:
+            branches:
+              only: client-python-release-branch
+      - release-approval:
+          type: approval
+          requires:
+            - release-github
+          filters:
+            branches:
+              only: client-python-release-branch
       - release-pypi:
+          requires:
+            - release-approval
           filters:
             branches:
               only: client-python-release-branch
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -113,7 +113,7 @@ jobs:
     steps:
       - install-bazel-linux-rbe
       - checkout
-      - run: bazel run //:deploy-pip -- pypi $PYPI_REPO_USERNAME $REPO_PYPI_PASSWORD
+      - run: bazel run //:deploy-pip -- test $PYPI_REPO_USERNAME $REPO_PYPI_PASSWORD
 
   release-cleanup:
     machine: true
--- a/BUILD
+++ b/BUILD
@@ -16,7 +16,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
-exports_files(["requirements.txt"])
+exports_files(["requirements.txt", "deployment.properties", "RELEASE_TEMPLATE.md"])
 
 load("@io_bazel_rules_python//python:python.bzl", "py_library", "py_test")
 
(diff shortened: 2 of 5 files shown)


.circleci/config.yml
```diff
@@ -145,7 +153,20 @@ workflows:
 
   client-python-release:
     jobs:
+      - release-github:
+          filters:
+            branches:
+              only: client-python-release-branch
+      - release-approval:
```
haikalpribadi: let's rename this to `release-deployment-approval` so we don't get confused with the other `release-approval` which comes before the whole release workflow.

BUILD
```diff
@@ -105,6 +107,14 @@ deploy_pip(
 )
 
 
+deploy_github(
+    name = "deploy-github",
+    targets = [],
+    release_description = "//:RELEASE_TEMPLATE.md",
+    deployment_properties = "//:deployment.properties",
```
haikalpribadi: we don't need to pass a whole file (`deployment.properties`) just to pass the address to the GitHub repo here. I think a string argument containing the repo address is enough and simpler.

RELEASE_TEMPLATE.md
```diff
@@ -0,0 +1,5 @@
+# Grakn Client Python
+
+{ Please add the release notes here }
+
+{ Please add instructions on how to install }
```
haikalpribadi: Can you also add the instructions to install in this template? It should change very rarely, right?

deployment.properties
```diff
@@ -0,0 +1,20 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+repo.github.organisation=graknlabs
+repo.github.repository=client-python
```
haikalpribadi: As per my comment above, I don't think we need a new `deployment.properties` file just to contain the repository address. What do you think?

RELEASE_TEMPLATE.md
vmax: @haikalpribadi I'm not sure I know how to best phrase it. The only thing that comes to mind:
"To install Grakn Client Python, run `pip install -U grakn-client`"

deployment.properties
vmax: @haikalpribadi `deployment.properties` is kind of standard way to pass properties. Do we want really want to refactor the rules and its usage across all `@graknlabs_bazel_distribution` to take in repo URLs as rule attributes?

.circleci/config.yml
vmax: @haikalpribadi addressed in https://github.com/graknlabs/client-python/pull/36/commits/224287ccbff9b509cb7a8ae4ac46d6eb0849dbd6

RELEASE_TEMPLATE.md
haikalpribadi: Just explain how you think best. We'll review it then. Just make sure you provide complete instructions and point them to the docs where necessary.

ps: what is `-U` for?

vmax: @haikalpribadi `-U` is `upgrade` which makes sure user always installs the latest Grakn (in case he already has one installed)

deployment.properties
haikalpribadi: well, no, we don't want to refactor everything for this. But now we have 2 deployment.properties for every repo (one from build-tools and one from root). And it gets a bit confusing. Does one override the other? I imagine `deployment.properties` should contain "global" values that are shared across all repos. But _"where deploy_github should deploy to"_ feels like a one-off argument which is very event-specific, right?

RELEASE_TEMPLATE.md
haikalpribadi: is that really the command we should tell users to do?

deployment.properties
haikalpribadi: What if for `deploy_github` we pass `orgnisation` and `repository` arguments? Then we don't need to introduce a new `deployment.properties` file for each repo, right? Once there are more variables we need to manage per repo, then we refactor again to move them over to a `deployment.properties` file because then it would be worth managing them in a new file. What do you think?

RELEASE_TEMPLATE.md
vmax: @haikalpribadi yep. is there something wrong with it? 🤔 
We can also include [this link](https://dev.grakn.ai/docs/client-api/python#installation) to docs

deployment.properties
vmax: @haikalpribadi 
>But "where deploy_github should deploy to" feels like a one-off argument which is very event-specific, right?

Does it mean that we should **not** read values from `deployment.properties` for `deploy_github` and _always_ take in argument? We also use `deploy_github` in `grakn-core` where value is currently read from `build-tools`'s `deployment.properties`.

RELEASE_TEMPLATE.md
vmax: @haikalpribadi addressed in https://github.com/graknlabs/client-python/pull/36/commits/7b4db5b595d6de3b39df4bb97297881cbeab0ce6

BUILD
haikalpribadi: it turns out we do need it

PR #39, opened 2019-04-03
//:test-integration now depends on @graknlabs_grakn_core//:assemble-mac-zip

vmax:

Test should unpack, start/stop Grakn distribution _within_ the test

* tests now depend on `@graknlabs_grakn_core//:assemble-mac-zip` via `data`
* tests now unpack/start/stop Grakn before executing

### DIFF
--- a/tests/integration/base.py
+++ b/tests/integration/base.py
@@ -61,11 +61,6 @@ class GraknServer(object):
     def __init__(self):
         self.__unpacked_dir = None
 
-    def _unpack(self):
-        self.__unpacked_dir = tempfile.mkdtemp(prefix='grakn')
-        with ZipFile(GraknServer.DISTRIBUTION_LOCATION) as zf:
-            zf.extractall(self.__unpacked_dir)
-
     def __enter__(self):
         if not self.__unpacked_dir:
             self._unpack()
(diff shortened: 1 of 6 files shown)


tests/integration/base.py
```diff
@@ -34,6 +40,56 @@ def __exit__(self, *args, **kwargs):
...
+    def _unpack(self):
+        self.__unpacked_dir = tempfile.mkdtemp(prefix='grakn')
+        with ZipFile(GraknDistribution.DISTRIBUTION_LOCATION) as zf:
+            zf.extractall(self.__unpacked_dir)
+
+    @property
+    def __grakn_root_directory(self):
+        return os.path.join(self.__unpacked_dir, GraknDistribution.DISTRIBUTION_ROOT_DIR)
+
+    def __enter__(self):
```
lolski: As we've discussed over Slack, let's combine `start` with `__enter__` and `stop` with `__exit__`. `__enter__` and `__exit__` are just one liners so there isn't much sense in splitting them

tests/integration/base.py
```diff
@@ -34,6 +40,56 @@ def __exit__(self, *args, **kwargs):
...
+        self.start()
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.stop()
+
+    def start(self):
```
lolski: What if we get rid of `__grakn_root_directory `:

```suggestion
    def start(self):
        ...
        sp.check_call([
            'grakn', 'server', 'start'
        ], cwd=os.path.join(self.__unpacked_dir, DISTRIBUTION_ROOT_DIR))
```

I think `@property` adds an unnecessary noise to the code, especially considering most of us here aren't Python programmers

tests/integration/base.py
```diff
@@ -34,6 +40,56 @@ def __exit__(self, *args, **kwargs):
...
+        if not self.__unpacked_dir:
+            self._unpack()
+        sp.check_call([
+            'grakn', 'server', 'start'
+        ], cwd=self.__grakn_root_directory)
```
lolski: here as well

tests/integration/base.py
```diff
@@ -34,6 +40,56 @@ def __exit__(self, *args, **kwargs):
...
+
+    def stop(self):
+        sp.check_call([
+            'grakn', 'server', 'stop'
+        ], cwd=self.__grakn_root_directory)
```
lolski: and here

PR #42, opened 2019-04-19
Support KGMS 1.5.2

lolski:

**NOTE**: This change is an internal change, the public API stays the same.

Currently, `client-python` only supports KGMS 1.4.3 and won't work with 1.5. This is because in 1.5 the Session / Transaction architecture were simplified and `username` and `password` needs to be correctly supplied when opening a session rather than transaction. This PR fixes that.

When opening a session, `username` and `password` are supplied in the `Session.Open.Req` gRPC message.


dependencies/graknlabs/dependencies.bzl
```diff
@@ -26,8 +26,9 @@ def graknlabs_build_tools():
     )
 
 def graknlabs_grakn_core():
+    # TODO: revert to graknlabs
```
flyingsilverfin: Comment needed still?

lolski: thanks! removed in https://github.com/graknlabs/client-python/pull/42/commits/c41f905dd76e35e1d746818aed30d1b01e0a0c30

PR #45, opened 2019-05-13
Close Transactions Properly

### LINKED ISSUE #44: Flakey test failing to frequently
https://circleci.com/gh/graknlabs/client-python/1042
https://circleci.com/gh/graknlabs/client-python/1043

flyingsilverfin:
Properly close transactions that was leading to failing client-python tests, and ensure that tests that are failing clean up properly rather than exiting without closing transactions. 

* Exhaust the gRPC iterator when the transaction is closed to ensure that `onCompleted()` is called on the server. Previously, we were emptying the input iterator and adding `None` to its queue, but this was not being consumed. Thus, the transactions was not being cleaned up on the server as the `transaction(stream)` RPC call was never completed cleanly.
* Utilise `addCleanup` to close transactions even if the tests fail - this is always executed regardless of test status.

Closes #44

### DIFF
--- a/grakn/service/Session/TransactionService.py
+++ b/grakn/service/Session/TransactionService.py
@@ -170,5 +170,7 @@ class Communicator(six.Iterator):
                 self._queue.queue.clear()
             self._queue.put(None)
             self._closed = True
+            # force exhaust the iterator so `onCompleted()` is called on the server
+            # note: next() was throwing an error sometimes, but would be preferable to a loop
             for empty in self._response_iterator:
                 pass
--- a/grakn/service/Session/TransactionService.py
+++ b/grakn/service/Session/TransactionService.py
@@ -155,7 +155,8 @@ class Communicator(six.Iterator):
             self._add_request(request)
             response = next(self._response_iterator)
         except Exception as e: # specialize into different gRPC exceptions?
-            # on any GRPC exception, close the stream
+            # invalidate this communicator, functionally this occurs automatically on exception (iterator not usable anymore)
+            self._closed = True
             raise GraknError("Server/network error: {0}\n\n generated from request: {1}".format(e, request))
 
         if response is None:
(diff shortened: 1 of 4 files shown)


grakn/service/Session/TransactionService.py
```diff
@@ -165,7 +165,12 @@ def send(self, request):
...
-        self._queue.put(None)
-        self._closed = True
+        if not self._closed:
+            with self._queue.mutex: # probably don't even need the mutex
+                self._queue.queue.clear()
+            self._queue.put(None)
+            self._closed = True
+            # force exhaust the iterator so `onCompleted()` is called on the server
+            # note: next() was throwing an error sometimes, but would be preferable to a loop
+            for empty in self._response_iterator:
```
flyingsilverfin: @vmax maybe you have an idea on how to do this better. `next(...)` was not working I think!

grakn/service/Session/TransactionService.py
```diff
@@ -165,7 +165,12 @@ def send(self, request):
...
-            self._queue.queue.clear()
-        self._queue.put(None)
-        self._closed = True
+        if not self._closed:
+            with self._queue.mutex: # probably don't even need the mutex
+                self._queue.queue.clear()
+            self._queue.put(None)
+            self._closed = True
+            # force exhaust the iterator so `onCompleted()` is called on the server
+            # note: next() was throwing an error sometimes, but would be preferable to a loop
```
lolski: can you copy/paste the error that is thrown by `next()` @flyingsilverfin ?

flyingsilverfin: just double checking it was this error not another one covering it up! Will do if it still fails

flyingsilverfin: Further investigation shows `next()` was giving "ERROR", but this was caused by a `StopIteration` exception that just needs to be caught. updated.

PR #73, opened 2019-10-23
Delete Queries return Void Answer Type, add Concept.is_deleted()

flyingsilverfin:
As of https://github.com/graknlabs/protocol/pull/18, we have decided in a slight paradigm shift with `delete` queries:

1. straight delete queries `match...; delete...;` only return a message rather than the halfway house of all deleted Concept IDs (this was always awkward, should either return a Concept -- hard because vertex deleted, or nothing. We have opted for nothing).
2. If you want to know what was deleted, this implies your behavior was a _retrieve_ followed by a _delete_. This retrieve should be performed explicltly by the user not implicitly by Grakn, ie. a `match...; get...;` followed by `delete` using Concept API, or using a separate `match...; delete;` query.

Also, new method `is_deleted()` exists on all `Concept` as part of the Concept API.

* Delete queries passed to the client now return `Void` answer type with a status message
* New Concept API method `is_deleted()` that checks if a concept has been deleted to facilitate usage of the new `delete` paradigm

### DIFF
--- a/tests/integration/base.py
+++ b/tests/integration/base.py
@@ -77,7 +77,6 @@ class GraknServer(object):
     def _unpack(self):
         self.__unpacked_dir = tempfile.mkdtemp(prefix='grakn')
         with ZipFile(GraknServer.DISTRIBUTION_LOCATION) as zf:
-            print(self.__unpacked_dir)
             zf.extractall(self.__unpacked_dir)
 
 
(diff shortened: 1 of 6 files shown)


tests/integration/base.py
```diff
@@ -77,6 +77,7 @@ def __exit__(self, exc_type, exc_val, exc_tb):
     def _unpack(self):
         self.__unpacked_dir = tempfile.mkdtemp(prefix='grakn')
         with ZipFile(GraknServer.DISTRIBUTION_LOCATION) as zf:
+            print(self.__unpacked_dir)
```
vmax: @flyingsilverfin is this needed still?

flyingsilverfin: ah good catch! will remove :)

PR #79, opened 2020-01-14
Upgrade gRPC to 1.24.1

vmax:

Update gRPC to a more recent version

- Upgrade PyPI dependency on gRPC to 1.24.1
- Bump `@graknlabs_build_tools`, `@graknlabs_protocol ` and `@graknlabs_grakn_core` versions

### DIFF
--- a/tests/integration/test_concept.py
+++ b/tests/integration/test_concept.py
@@ -163,6 +163,8 @@ class test_SchemaConcept(test_concept_Base):
             # set label
             car_type = self.tx.get_schema_concept("car")
             car_type.label("vehicle")
+            # retrieve updated schema concepts
+            list(self.tx.query('match $x sub thing; get;'))
             vehicle_type = self.tx.get_schema_concept("vehicle")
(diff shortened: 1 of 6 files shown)


tests/integration/test_concept.py
```diff
@@ -163,6 +163,8 @@ def test_set_label(self):
             # set label
             car_type = self.tx.get_schema_concept("car")
             car_type.label("vehicle")
+            # retrieve updated schema concepts
```
flyingsilverfin: why is this needed @vmax?

vmax: @flyingsilverfin
```
q = tx.put_entity_type('q')
tx.get_schema_concept('q') # works
q.label('w')

tx.get_schema_concept('q') # no longer works (expected)
tx.get_schema_concept('w') # ALSO doesn't work
```

however:

```
tx.get_schema_concept('w')  # None
list(tx.query('match $x sub thing; get;')) # list<conceptmap>
tx.get_schema_concept('w') # it is there!
```

which leads me to believe schema concepts might be cached somewhere

PR #88, opened 2020-04-15
Upgrade client-python to new protocol for grakn 1.7

adammitchelldev:

Update client-python to the new 1.0.5 protocol with changed iteration style and local/remote concept API split.

The new protocol significantly reduces the number of round trips necessary to complete queries, down to 1 in most cases, from the previous solution which required some multiple of the number of results. This should alleviate slowdown problems caused by connecting to a remote Grakn instance. Iteration now follows a streaming principle and results are pre-filled with data the user would normally be expected to retrieve via the concept API.

The step to split local and remote concepts helps us move towards "value" results rather than results which need to embed the transaction they were retrieved with. The added requirement to convert these concepts back to remote concepts should also make it clear to the user which methods require additional RPCs (where it wasn't clear beforehand that ALL methods required additional RPCs).

* Implement local and remote concepts.
* Use local concept for conceptmap answers.
* Implement new iteration for queries, get_attributes_by_value and all iterating concept methods


grakn/service/Session/Concept/Concept.py
```diff
@@ -18,610 +18,180 @@
...
-        method_response = self._tx_service.run_concept_method(self.id, del_request)
-        return
-
-    def is_deleted(self):
-        retrieved = self._tx_service.get_concept(self.id)
-        return retrieved is None
+    def __init__(self, grpc_concept):
+        self.id = grpc_concept.id
 
+    def as_remote(self, tx_service):
```
flyingsilverfin, line 32: why does the user have a handle to the tx_service? Should this be the plain `tx`?

grakn/service/Session/Concept/ConceptFactory.py
```diff
@@ -17,34 +17,52 @@
 # under the License.
 #
 
-from grakn.service.Session.Concept import BaseTypeMapping
-from grakn.service.Session.Concept.Concept import EntityType, RelationType, AttributeType, Role, Rule, Entity, Relation, Attribute, Type
-
+#from grakn.service.Session.Concept import BaseTypeMapping
+import BaseTypeMapping
```
vmax, line 21: @adammitchelldev this import doesn't quite look fine to me :thinking: 
have you tested it?

grakn/service/Session/Concept/ConceptFactory.py
```diff
@@ -17,34 +17,52 @@
...
+import Concept
+#from grakn.service.Session.Concept.Concept import name_to_object, EntityType, RelationType, AttributeType, Role, Rule, Entity, Relation, Attribute, Type
```
flyingsilverfin: cleanup imports

grakn/service/Session/Concept/ConceptFactory.py
adammitchelldev: Yes, it does work fine but it can be reverted now I think.

grakn/service/Session/TransactionService.py
```diff
@@ -39,16 +40,11 @@ def __init__(self, session_id, tx_type, transaction_endpoint):
...
     # --- Passthrough targets ---
     # targets of top level Transaction class
 
     def query(self, query, infer=True):
-        request = RequestBuilder.query(query, infer=infer)
-        # print("Query request: {0}".format(request))
-        response = self._communicator.send(request)
-        # convert `response` into a python iterator
-        return ResponseReader.ResponseReader.query(self, response.query_iter)
+        return imap(ResponseReader.ResponseReader.query(self), Iterator(self._communicator, RequestBuilder.start_iterating_query(query, infer)))
```
flyingsilverfin, line 47: can you write a comment on what this imap does here?

grakn/service/Session/TransactionService.py
```diff
@@ -117,17 +113,61 @@ def run_concept_method(self, concept_id, grpc_concept_method_req):
...
-        return response.iterate_res
-    iterate.__annotations__ = {'iterator_id': int}
+
+class Iterator(six.Iterator):
+    def __init__(self, communicator, iter_req):
+        self._id = 0
+        self._communicator = communicator
+        self._iter_req = iter_req
+        self._buffer = []
+        self._start_iterating()
```
flyingsilverfin, line 132: Is it intended that we start iterating right away? We aren't lazy anymore then right?

grakn/service/Session/TransactionService.py
```diff
@@ -152,6 +192,15 @@ def __next__(self):
...
+    def send_only(self, request):
+        if self._closed:
+            raise GraknError("This connection is closed")
+        try:
+            self._add_request(request)
+        except Exception as e:
+            self._closed = True
+            raise GraknError("Server/network error: {0}\n\n generated from request: {1}".format(e, request))
+
     def send(self, request):
```
flyingsilverfin, line 204: I think renaming to `send_receive`  is a good idea

grakn/service/Session/util/RequestBuilder.py
```diff
@@ -176,17 +208,15 @@ def explanation(explainable):
...
-    @staticmethod
-    def next_iter(iterator_id):
-        iterate_request = transaction_messages.Transaction.Iter.Req()
-        iterate_request.id = iterator_id
-        
-        transaction_req = transaction_messages.Transaction.Req()
-        transaction_req.iterate_req.CopyFrom(iterate_request)
-        return transaction_req
-
+    #
```
flyingsilverfin, line 211: leftover commented code

grakn/service/Session/util/ResponseReader.py
```diff
@@ -294,7 +300,7 @@ def _create_concept_map(tx_service, grpc_concept_map_msg):
     @staticmethod
     def _create_answer_group(tx_service, grpc_answer_group):
         grpc_owner_concept = grpc_answer_group.owner
-        owner_concept = ConceptFactory.create_concept(tx_service, grpc_owner_concept)
+        owner_concept = ConceptFactory.create_remote_concept(tx_service, grpc_owner_concept)
```
flyingsilverfin, line 303: do we require remote concepts here?

grakn/service/Session/Concept/Concept.py
adammitchelldev: It actually is the plain `tx` which is one of the bugs I had to fix, will rename it now.

haikalpribadi:
Might want to rename the PR title properly and elaborate on the description, though, @adammitchelldev . They will be used for the release notes.

grakn/service/Session/TransactionService.py
adammitchelldev: It's mapping the response reader to each result of the iterator, is the comment definitely necessary?

grakn/service/Session/TransactionService.py
adammitchelldev: If we don't immediately send the query then define and insert queries, which are not expected to have results, will not get executed (since currently the iterators are not executed).

We can choose to use a batch size of 0 for the first iteration, but that would introduce an extra round trip to gets where we are actually expecting a result.

Also, in general, being lazy is undesirable on the network, since latency is nearly always part of the critical path of the operation (or at least, should be after optimisation).

grakn/service/Session/TransactionService.py
adammitchelldev: Agreed

grakn/service/Session/util/ResponseReader.py
adammitchelldev: I'm not sure but I haven't done any analysis on what might be necessary to pre-fetch here, so this is how we currently do it in Java.

grakn/service/Session/util/RequestBuilder.py
adammitchelldev: > Might want to rename the PR title properly and elaborate on the description, though, @adammitchelldev . They will be used for the release notes.

Gah, I swear I did!

grakn/service/Session/TransactionService.py
flyingsilverfin: ah yeah i forgot about define and insert queries, this is important.

grakn/service/Session/util/ResponseReader.py
flyingsilverfin: ok, can you try using a local concept and see if it breaks any tests? It shouldn't need any remote functionality...

flyingsilverfin:
To be clear, this isn't doing the asynchronous refiling of a shared buffer right? Ie. the channel will be blocked after calling the batch iterator until the batch is received?

flyingsilverfin:
Also I think you could improve description of the goal to include "Reduce round trips and read query latency" or similar, so people see it in the release notes :)

grakn/service/Session/TransactionService.py
flyingsilverfin: It's just hard to read what is intended to be going on here: are we mappingthe query() function with new arguments generated by the iterator? Does that mean the `query(self)` method returns a function that takes further arguments?

Functional style programming is great but i find it hard to follow when mixed in with other styles :)

adammitchelldev: Moved the `imap` to `ResponseReader` to be consistent with the style for `get_attributes_by_value`. This method previously existed and was called `query` but returned an iterator that it created, since iteration was handled by `ResponseReader` prior. I have also renamed the method to `get_query_results` now, and it takes an iterator as as parameter rather than being mapped (so that the `imap` and the `lambda` are in the same place).

grakn/service/Session/util/ResponseReader.py
```diff
@@ -27,19 +28,15 @@
 class ResponseReader(object):
     
     @staticmethod
-    def query(tx_service, grpc_query_iter):
-        iterator_id = grpc_query_iter.id
-        return ResponseIterator(tx_service,
-                                iterator_id,
-                                lambda tx_serv, iterate_res: AnswerConverter.convert(tx_serv, iterate_res.query_iter_res.answer))
-
+    def query(tx_service):
```
flyingsilverfin, line 31: Ok so i think this method's naming is the incorrect - this is really the `ReasonerReader.from_iterate_response` or something right? `query` is too generic and doesn't carry any meaning anymore

grakn/service/Session/TransactionService.py
flyingsilverfin: i think if you rename `query(self)` above like i mentioned in other comment it will help a lot, right now this just isn't readable standalone :)

PR #91, opened 2020-05-12
Relax strict dependencies on external packages

vmax:

Some of our users reported issues on external dependencies being too strict.

Relax dependencies such that any versions of `grpc` (as long as major is less that `2`) and `enum-compat` could be used

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,6 +45,7 @@ jobs:
     steps:
       - install-bazel-linux-rbe
       - checkout
+      - run: pyenv global 2.7.12 3.5.2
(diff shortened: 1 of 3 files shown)


flyingsilverfin:
@vmax  its the enum lib not the grpc lib that is failing right?

vmax:
@flyingsilverfin I suspect this is connected to us using Python 2 to do tests: `enum` is not available there

flyingsilverfin:
any ideas on what we can do then? Im surprised using enum 0.0.2 or 0.0.3 isn't compatible with each other...

PR #92, opened 2020-05-12
Fix batch size to 50

adammitchelldev:

To set the query batch size to 50 by default, rather than all. This was a mistake when releasing, the optional batch size in the RequestBuilder was intended to be used by the query method in the query thing. The batch size should also be optional for users who require small batches.

- Batch size option available on queries, default value is 50.


flyingsilverfin:
We can close this right @adammitchelldev ?

PR #93, opened 2020-05-12
Add async iteration

adammitchelldev:

To add async iteration to `client-python` so that results can be iterated whilst still waiting to be returned.

This implementation follows the architecture found in `client-nodejs`. When requests are made, they are immediately sent to the server and at the same time, a "response resolver" is placed onto the "resolver" queue. The response resolver queue ensures that responses are handled in the order that the requests are sent, thus allowing multiple requests to be sent whilst the responses are still being resolved, which is what enables us to handle async correctly.

When attempting to retrieve a response, the GRPC stream is iterated and the responses are buffered into each resolver until the desired response is found and can be returned. This ensures that messages are not lost due to out-of-order iteration.

- Added async iteration
- Queries are now fully async, so multiple queries can be made without additional roundtrips (until results are used).


grakn/service/Session/util/Communicator.py
```diff
@@ -0,0 +1,112 @@
...
+#
+
+import six
+from collections import deque
+from six.moves import queue
+from grakn.exception.GraknError import GraknError
+
+
+class IterationResolver(six.Iterator):
+    def __init__(self, communicator, request, is_last_response=lambda x: True):
```
flyingsilverfin, line 27: can we not have a default for the last arg so it's always required here?

flyingsilverfin:
It's not clear to me the threading model (if there is one) going on here, ill take a second look but maybe we can both add a descrption about how it works to the PR and in the code somewhere

grakn/service/Session/TransactionService.py
```diff
@@ -36,19 +37,19 @@ def __init__(self, session_id, tx_type, transaction_endpoint):
...
+        self._communicator.single_request(open_req)
     __init__.__annotations__ = {'tx_type': enums.TxType}
 
     # --- Passthrough targets ---
     # targets of top level Transaction class
 
-    def query(self, query, infer=True):
-        return ResponseReader.ResponseReader.get_query_results(self, Iterator(self._communicator, RequestBuilder.start_iterating_query(query, infer)))
+    def query(self, query, infer=True, batch_size=50):
+        return ResponseReader.ResponseReader.get_query_results(self, Iterator(self._communicator, RequestBuilder.start_iterating_query(query, infer, batch_size)))
```
flyingsilverfin, line 47: can we split this line out a little? Starting to get hard to read!

grakn/service/Session/util/Communicator.py
```diff
@@ -0,0 +1,112 @@
...
+                current._buffer_response(response)  # This result is for another
+
+            except Exception as e:
+                self._closed = True
+                if not current:
+                    raise GraknError("Internal client/protocol error, request/response pair not matched: {0}\n\n Ensure client version is compatible with server version.".format(e))
+                raise GraknError("Server/network error: {0}\n\n generated from request: {1}".format(e, current._request))
+
+    def iteration_request(self, request, is_last_response):
+        return IterationResolver(self, request, is_last_response)
```
flyingsilverfin, line 97: I think that this method shouldn't live in the `Communicator` - it obfuscates the flow of control and data. The callers of `iteration_request` can directly instantiate `IterationResolver(communicator, request, is_last_response` and we avoid having super-circular code to follow (iterator -> communicator.iteration_request -> IterationResolver -> communicator._send_with_resolver), it could look like (iterator -> IterationResolver -> communicator) instead...

the only thing with this is im not sure where to put `single_request` as it's used in multiple places... sure we can figure something out though :)

grakn/service/Session/util/Communicator.py
adammitchelldev: Hmm, I had it like that before but this felt a bit more concise, will revert

grakn/service/Session/TransactionService.py
adammitchelldev: Sure

grakn/service/Session/util/Communicator.py
adammitchelldev: Hmm, I'm not sure I prefer this.

The circularity is rather inherent since there are two "directions" that the messages come from:
A) pulled by requesting a response from an `IterationResolver`
B) pushed to an `IterationResolver` by the `Communicator` when a later resolver needs to pull a response

We could avoid any circular methods at all by putting all of the methods on `Communicator`, but that is also not necessarily nice to read (and the `IterationResolver` iterations methods are required so that it can be used as an iterator by the client).

The way I think of `IterationResolver` is as more of an implementation detail of `Communicator` and not a public class, since `Communicator` relies heavily on the hidden parts of `IterationResolver` to work correctly, they are collaborator classes at the very least but `Communicator` is the "owner". `IterationResolver` also happens to be the client interface to keep things simple, since that way there is only one object to worry about per request.

The fact that it is hard to read is really just purely inherent to the problem we are trying to solve, but we abstract it by making our API as simple and understandable to the user class (`TransactionService`) as possible:
* Single request calls return a response
* Multi request calls return an iterator

Changing that would obfuscate what we want the communicator to achieve as a user.

adammitchelldev: As an additional point, whilst the circularity of the code flow is harder to follow, it makes sense to me that the only "public" methods are the request methods and the `__next__` method on the `IterationResolver`, so those are the only places you need to start.

adammitchelldev:
> It's not clear to me the threading model (if there is one) going on here, ill take a second look but maybe we can both add a descrption about how it works to the PR and in the code somewhere

The threading model is: there are no threads. Everything is happening on a single thread, with the only exception being the request queue, which is required to work using a concurrent blocking queue. (This is the actual `Communicator` iterator, it puts the requests into the `six.queue` and the grpc stream iterates it, blocking each time until the next request is sent.)

When a request is made, the "receiver" of the request (called a `Resolver`, since it does not listen but instead pulls and needs to resolve whether or not the message is terminating) is added to a resolver queue and the request itself is immediately sent.

Nothing happens until a response is requested through a `Resolver`.

When it is, the `Resolver` first checks if its message was already processed (the buffered messages) and returns that. This only happens if another `Resolver` which was placed later was requested first, messages are never pushed via the GRPC, they get pulled by the client thread to make way for another resolver.

If the `Resolver` does not already have a message in the buffer to return (and is not empty), it must iterate all resolvers in the queue, pushing the responses as they are received from the GRPC stream (which it pulls from) to the corresponding resolver buffers until it can acquire a new message for the current resolver.

If no resolvers are called out of order, each resolver will simply pull one message from the communicator's GRPC stream with no extra threading.

grakn/service/Session/util/Communicator.py
flyingsilverfin: Ok - it was just unclear to me whether the `Communicator` or the `IterationResolver` has control in this setup.

the other solution to retain the simple API you've made is to introduce a small class/layer that does iteration_request() and single_request() but i'm not sure that would make the code any easier to read

flyingsilverfin: I don't mind the circularity at a concept, just push it down a bit further so that the first thing the communicator does isn't create a `IterationResolver` which just calls the communicator again - the entry point could just be one directional (and the cirularity occurs deeper down)

grakn/service/Session/TransactionService.py
```diff
@@ -109,136 +112,62 @@ def put_rule(self, label, when, then):
...
-
-    def __init__(self, grpc_stream_constructor):
-        self._queue = queue.Queue()
-        self._response_iterator = grpc_stream_constructor(self)
-        self._closed = False
-
-    def _add_request(self, request):
-        self._queue.put(request)
+    def _request_next_batch(self, iter_id):
+        self._response_iterator = self._communicator.iteration_request(RequestBuilder.iter_req_to_tx_req(
```
flyingsilverfin, line 150: i think this would also read a bit easier if split out :P

grakn/service/Session/util/Communicator.py
```diff
@@ -0,0 +1,143 @@
...
+# under the License.
+#
+
+import six
+from collections import deque
+from six.moves import queue
+from grakn.exception.GraknError import GraknError
+
+
+class SingleResolver:
```
flyingsilverfin, line 26: nice, looks a lot more consistent!

grakn/service/Session/util/Communicator.py
```diff
@@ -0,0 +1,143 @@
...
+    def error_if_closed(self):
+        if self._closed:
+            raise GraknError("This connection is closed")
+
+    # Put a request for GRPC to consume
+    def _send_with_resolver(self, resolver, request):
+        self._resolver_queue.append(resolver)
+        self._request_queue.put(request)
+
+    def _block_for_next(self, resolver):
```
flyingsilverfin, line 102: since i think the meat of the demultiplexing of the stream of responses happens here, would be good to write a docstring with a small explanation.

side note, i think way you've got the "parallel" lists of request/resolver lists to keep track of which resolver to give answers, taking advantage of the gRPC ordering guarantees is neat!

adammitchelldev: I added it at the top of the file, I think that *should* suffice, what do you think?

flyingsilverfin: yep

PR #95, opened 2020-05-15
Don't use Python 2 in sync-dependencies

vmax:

#94 introduced running `sync-dependencies` with Python 2. Apparently, it's not compatible so we revert the behaviour.

Remove setting Python 2 as default for `sync-dependencies-snapshot`


flyingsilverfin:
shooot forgot to copy across PR descrption lol nvm

PR #105, opened 2020-06-09
Add future-style get for explicit waiting and error handling

adammitchelldev:

Since the introduction of asynchronous query processing, the error handling model has become less clear, as an error could be picked up on a line unrelated to its corresponding query. In order to allow clients to explicitly consume query completion, a `get()` method is added to the query result (iterator) which will block until the results are received, or throw an exception on error.

Clients looking to benefit from the asynchronous processing can continue without using the `get()` syntax.

- Added a `get()` method on the iterator to block until it has received either positive results or an error.

### DIFF
--- a/dependencies/graknlabs/dependencies.bzl
+++ b/dependencies/graknlabs/dependencies.bzl
@@ -27,8 +27,8 @@ def graknlabs_build_tools():
 def graknlabs_grakn_core():
     git_repository(
         name = "graknlabs_grakn_core",
-        remote = "https://github.com/flyingsilverfin/grakn",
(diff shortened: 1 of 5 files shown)


adammitchelldev:
In this implementation, I deliberately didn't remove the old iterator style syntax, I actually think this syntax is cleaner despite the error handling, because it has occurred to me that once one error has occurred, **no** further operations on the transaction should be considered valid. Bear in mind that this is for actual errors, which likely indicate a syntax issue in the clients' queries or a fault in grakn; currently errors kill the transaction because they throw a terminating error on the RPC.

Instead of relying on the `get()` style to sync error handling, we could consider the error handling only as a "total failure" of the transaction (any method can throw once the failure has happened), even if it does not correlate to a success.

adammitchelldev:
Closing due to CI issue

adammitchelldev:
Re-raised in #109

PR #108, opened 2020-06-12
Add Explanation.get_rule()

flyingsilverfin:
The `Rule` that corresponds to an explanation is now being returned in the protocol responses, but the python `Explanation` object does not record it. This PR records it if the rule is valid, else sets it to None.

* Add `Explanation.get_rule()`, and record the rule returned from the protocol
* Update a test to retrieve the rule as well

### DIFF
--- a/grakn/service/Session/util/ResponseReader.py
+++ b/grakn/service/Session/util/ResponseReader.py
@@ -127,18 +127,26 @@ class ResponseReader(object):
         native_list_of_concept_maps = []
         for grpc_concept_map in grpc_list_of_concept_maps:
             native_list_of_concept_maps.append(AnswerConverter._create_concept_map(tx_service, grpc_concept_map))
-        return Explanation(native_list_of_concept_maps)
+        rule = ConceptFactory.create_local_concept(grpc_explanation_res.rule)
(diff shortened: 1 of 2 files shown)


grakn/service/Session/util/ResponseReader.py
```diff
@@ -127,19 +127,27 @@ def create_explanation(tx_service, grpc_explanation_res):
         native_list_of_concept_maps = []
         for grpc_concept_map in grpc_list_of_concept_maps:
             native_list_of_concept_maps.append(AnswerConverter._create_concept_map(tx_service, grpc_concept_map))
-        return Explanation(native_list_of_concept_maps)
+        rule = ConceptFactory.create_local_concept(grpc_explanation_res.rule)
+        if len(rule.id) == 0:
```
vmax, line 131: @flyingsilverfin 

```suggestion
        if not rule.id:
```

flyingsilverfin: i like the explicit length based version as it doesn't require any specialised python knowledge about strings and not behaviour, easy to read for people swapping between langauges a lot :)

PR #109, opened 2020-06-15
Add future-style get for explicit waiting and error handling

adammitchelldev:

Since the introduction of asynchronous query processing, the error handling model has become less clear, as an error could be picked up on a line unrelated to its corresponding query. In order to allow clients to explicitly consume query completion, a `get()` method is added to the query result (iterator) which will block until the results are received, or throw an exception on error.

Clients looking to benefit from the asynchronous processing can continue without using the `get()` syntax.

- Added a `get()` method on the iterator to block until it has received either positive results or an error.

### DIFF
--- a/grakn/service/Session/TransactionService.py
+++ b/grakn/service/Session/TransactionService.py
@@ -44,9 +44,9 @@ class TransactionService(object):
     # targets of top level Transaction class
 
     def query(self, query, infer=True, batch_size=50):
-        return ResponseReader.ResponseReader\
-            .get_query_results(self, Iterator(self._communicator,
-                                              RequestBuilder.start_iterating_query(query, infer, batch_size)))
+        return Iterator(self._communicator,
+                        RequestBuilder.start_iterating_query(query, infer, batch_size),
+                        ResponseReader.ResponseReader.get_query_results(self))
     query.__annotations__ = {'query': str}
 
     def commit(self):
@@ -73,9 +73,8 @@ class TransactionService(object):
 
     def get_attributes_by_value(self, attribute_value, value_type):
         request = RequestBuilder.start_iterating_get_attributes_by_value(attribute_value, value_type)
-        iterator = Iterator(self._communicator, request)
-        return ResponseReader.ResponseReader.get_attributes_by_value(self, iterator)
-    get_attributes_by_value.__annotations__ = {'value_type': enums.ValueType}
+        return Iterator(self._communicator, request, ResponseReader.ResponseReader.get_attributes_by_value(self))
+    get_attributes_by_value.__annotations__ = {'data_type': enums.ValueType}
 
     def put_entity_type(self, label):
         request = RequestBuilder.put_entity_type(label)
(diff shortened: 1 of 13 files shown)


grakn/service/Session/util/Communicator.py
```diff
@@ -85,13 +85,15 @@ def __init__(self, communicator, request, is_last_response):
...
         self._ended = False
+        self._started = False
         self._response_buffer = deque()
         communicator._send_with_resolver(self, request)
 
     def __iter__(self):
         return self
 
     def __next__(self):
+        self._started = True
```
flyingsilverfin, line 96: is this required if it's also in `_buffer_response()`?

grakn/service/Session/TransactionService.py
```diff
@@ -173,4 +173,9 @@ def __next__(self):
             self._request_next_batch(iter_res.iteratorId)
             return next(self)
         else:
-            return iter_res
+            return self._resp_converter(iter_res)
+
+    def get(self):
```
flyingsilverfin, line 178: comment saying: blocks until the next response?

==> inclear if this can be used to force blocking with `next()` so if i wanted to force blocking i could do this:
``` 
iter.get()
for ans in iter:
  iter.get()
```
or somethign?

tests/integration/test_grakn.py
```diff
@@ -217,6 +217,17 @@ def test_no_metatype_duplicates(self):
...
+        answer = self.tx.query("match $x isa unicorn; get;")
+
+        with self.assertRaises(GraknError):
+            answer.get()
+
+    def test_query_errors_async_on_commit(self):
+        self.tx.query("match $x isa unicorn; get;")
+
+        with self.assertRaises(GraknError):
+            self.tx.commit()
```
flyingsilverfin, line 230: some tests on `.get()` blocking would be nice i think, if you can think of a way to do it :)

grakn/service/Session/TransactionService.py
adammitchelldev: This should do what you are expecting, but:
```
for ans in iter:
  pass
```
Logically does the same thing. The blocking may be useful if you want to prevent some complex computation taking place until a result has been received, or to ensure that there are otherwise no errors before executing something else.

grakn/service/Session/util/Communicator.py
adammitchelldev: Yes, `_buffer_response()` is called when the result is received before the client attempts to get it. `__next__()` may be called first, in which case it blocks and a corresponding call to `_buffer_response()` is not made.

adammitchelldev: To be more accurate, `_buffer_response()` is actually called when you attempt to receive a result that comes after the result of a previous request, and so you must buffer those results to the corresponding resolver.

tests/integration/test_grakn.py
adammitchelldev: I can't, not reliably in the current instrumentation anyway. I think we really need to revisit this at some later point as a client architecture revamp because there currently isn't much purpose in blocking.

PR #119, opened 2020-07-14
Depend on artifact distribution for tests

adammitchelldev:

Client-java does not depend on grakn core to build, only to test. By depending on the new artifact distribution, we can eliminate many bazel workspace dependencies that may cause consistency issues during tests.

We also need to change the CI to run jobs in parallel because right now they are unable to run the test suite without relying on RBE to provide isolation for each sub-test.

- Removed grakn core dependencies
- Updated to new dependencies form
- Depend on grakn core as artifact
- Rework test framework to use the artifact form
- Split tests in CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,7 +47,7 @@ jobs:
       - run-bazel:
           command: bazel build //...
 
-  test:
+  test-concept:
     machine: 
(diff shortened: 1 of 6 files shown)


adammitchelldev:
This PR may not be mergeable yet.

PR #130, opened 2020-11-26
Test using latest Python in Grabl

alexjpwalker:

To investigate whether it's possible to build the current client-python in Grabl with the Python version installed in Grabl.


alexjpwalker:
This PR should be deleted once Max's PR (that it is based off) is merged.

PR #134, opened 2020-12-09
Build remotely on Ubuntu 20.04

vmax:

In order to speed up our builds, we should do them remotely on BuildBuddy

Tweak `.bazelrc` settings to build on BuildBuddy with our own Ubuntu 20.04 image


lriuui0x0:
@vmax @alexjpwalker Is client-python build still not working?

vmax:
Replaced with #167

PR #143, opened 2020-12-29
Fix session pulses never getting sent

alexjpwalker:

The existing implementation of session renewal was not working, at all. Sessions would simply time out after 10 seconds. We fixed it.

Fix session pulses never getting sent

### DIFF
--- a/grakn/rpc/session.py
+++ b/grakn/rpc/session.py
@@ -36,6 +36,8 @@ class SessionType(enum.Enum):
 
 class Session(object):
 
+    _PULSE_FREQUENCY_SECONDS = 5
+
     def __init__(self, client, database: str, session_type: SessionType, options=GraknOptions()):
         self._channel = client._channel
         self._scheduler = sched.scheduler(time.time, time.sleep)
@@ -50,10 +52,10 @@ class Session(object):
 
         self._session_id = self._grpc_stub.session_open(open_req).session_id
         self._is_open = True
-        self._scheduler.enter(5, 1, self._transmit_pulse, ())
-        thread = Thread(target=self._scheduler.run)
-        # TODO: We should probably kill this thread when the session closes as it prevents the process closing.
-        thread.start()
+        self._scheduler.enter(delay=self._PULSE_FREQUENCY_SECONDS, priority=1, action=self._transmit_pulse, argument=())
+        # TODO: This thread blocks the process from closing. We should try cancelling the scheduled task when the
+        #       session closes. If that doesn't work, we need some other way of getting the thread to exit.
+        Thread(target=self._scheduler.run).start()
 
     def transaction(self, transaction_type: TransactionType, options=GraknOptions()):
         return Transaction(self._channel, self._session_id, transaction_type, options)
@@ -78,9 +80,8 @@ class Session(object):
         pulse_req.session_id = self._session_id
         res = self._grpc_stub.session_pulse(pulse_req)
         if res.alive:
-            self._scheduler.enter(5, 1, self._transmit_pulse, ())
-            thread = Thread(target=self._scheduler.run)
-            thread.start()
+            self._scheduler.enter(delay=self._PULSE_FREQUENCY_SECONDS, priority=1, action=self._transmit_pulse, argument=())
+            Thread(target=self._scheduler.run).start()
 
     def __enter__(self):
         return self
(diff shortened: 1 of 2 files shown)


grakn/rpc/session.py
```diff
@@ -70,9 +75,11 @@ def _transmit_pulse(self):
             return
         pulse_req = session_proto.Session.Pulse.Req()
         pulse_req.session_id = self._session_id
-        is_alive = self._grpc_stub.session_pulse(pulse_req).is_alive
-        if is_alive:
-            self._pulse = self._scheduler.enter(5, 1, self._transmit_pulse, ())
+        res = self._grpc_stub.session_pulse(pulse_req)
+        if res.alive:
```
alexjpwalker, line 79: Fixing the pulse scheduler also exposed another bug: the field was named wrongly (was `is_alive`, should be `alive`)

grakn/rpc/session.py
```diff
@@ -47,7 +50,9 @@ def __init__(self, client, database: str, session_type: SessionType, options=Gra
 
         self._session_id = self._grpc_stub.session_open(open_req).session_id
         self._is_open = True
-        self._pulse = self._scheduler.enter(5, 1, self._transmit_pulse, ())
+        self._scheduler.enter(5, 1, self._transmit_pulse, ())
+        thread = Thread(target=self._scheduler.run)
+        thread.start()
```
alexjpwalker, line 55: The reason pulses were never triggered was because we were never calling `self._scheduler.run`.

alexjpwalker (commented):
These changes look good.

grakn/rpc/session.py
```diff
@@ -47,7 +50,10 @@ def __init__(self, client, database: str, session_type: SessionType, options=Gra
 
         self._session_id = self._grpc_stub.session_open(open_req).session_id
         self._is_open = True
-        self._pulse = self._scheduler.enter(5, 1, self._transmit_pulse, ())
+        self._scheduler.enter(5, 1, self._transmit_pulse, ())
+        thread = Thread(target=self._scheduler.run)
+        # TODO: We should probably kill this thread when the session closes as it prevents the process closing.
```
vmax, line 55: @alexjpwalker I don't think we can _kill_ thread in Python - it can only stop itself (by ceasing to run), see [ref](https://stackoverflow.com/questions/323972/is-there-any-way-to-kill-a-thread)

grakn/rpc/session.py
```diff
@@ -47,7 +50,10 @@ def __init__(self, client, database: str, session_type: SessionType, options=Gra
...
+        thread.start()
```
vmax, line 56: @alexjpwalker given we only use `thread` in this place, we could inline it like so:

```suggestion
        Thread(target=self._scheduler.run).start()
```

grakn/rpc/session.py
```diff
@@ -47,7 +50,10 @@ def __init__(self, client, database: str, session_type: SessionType, options=Gra
 
         self._session_id = self._grpc_stub.session_open(open_req).session_id
         self._is_open = True
-        self._pulse = self._scheduler.enter(5, 1, self._transmit_pulse, ())
+        self._scheduler.enter(5, 1, self._transmit_pulse, ())
```
vmax, line 53: @alexjpwalker `5` and `1` look awfully like magic numbers here :slightly_smiling_face: the way to improve it would be to name arguments (so we don't need to look into docs to understand them), separate values into constants (to avoid magic numbers) and avoid passing optional arguments (`argument` is only optional since 3.3 so it really depends on minimum Python we support)

```suggestion
        self._scheduler.enter(delay=SESSION_PULSE_DELAY, priority=SESSION_PULSE_PRIORITY, action=self._transmit_pulse)
```

vmax (approved):
LGTM, left some suggestions

grakn/rpc/session.py
alexjpwalker: Ah I see. Would cancelling all the scheduled tasks cause the thread to close? I think it would - will update the comment accordingly

grakn/rpc/session.py
alexjpwalker: Updated to `self._scheduler.enter(delay=self._PULSE_FREQUENCY_SECONDS, priority=1, action=self._transmit_pulse, argument=())`. (We probably won't officially support older Pythons but I see no reason to _intentionally_ break compatibility)

PR #147, opened 2021-01-05
Delete query_pattern from ConceptMap

alexjpwalker:

Delete query_pattern from ConceptMap


Phillammon (approved):
Appears correct, seems consistent with other clients. Approved.

PR #148, opened 2021-01-07
RPC implementation for sending match aggregate, match group, and match group aggregate queries

lolski:

We have implemented RPC support for sending match aggregate, match group, and match group aggregate queries to server

1. Update `@graknlabs_protocol` to the version that supports the new queries.
2. Support processing the new queries


tests/integration/test_query.py
```diff
@@ -53,8 +53,8 @@ def test_insert_some_entities(self):
         with client.session("grakn", SessionType.DATA) as session:
             with session.transaction(TransactionType.WRITE) as tx:
                 for answer in tx.query().insert("insert $a isa lion; $b isa lion; $c isa lion;"):
-                    print(answer)
-
+                    print("insert: " + str(answer))
+                    tx.commit()
```
lolski, line 57: Spotted a write transaction where we forgot to commit

tests/integration/test_grakn.py
```diff
@@ -23,7 +23,7 @@
 from grakn import GraknClient, ValueType, Transaction
 from grakn import GraknError
 
-from test.integration.base import test_base, GraknServer
+from tests.integration.base import test_base, GraknServer
```
flyingsilverfin, line 26: any particular reason for this @lolski ?

grakn/concept/answer/concept_map.py
```diff
@@ -23,12 +23,12 @@
...
 from grakn.common.exception import GraknClientException
 from grakn.concept.proto import concept_proto_reader
-from grakn.concept.answer.answer import Answer
 from grakn.concept.concept import Concept
 
 
-class ConceptMap(Answer):
+class ConceptMap:
 
+    _CONCEPT_MAP = "concept_map"
```
flyingsilverfin, line 31: i don't think you're actually using this anywhere

grakn/query/query_manager.py
```diff
@@ -17,13 +17,13 @@
 # under the License.
 #
 
-from typing import Callable, List
+from typing import Callable, List, Any
```
flyingsilverfin, line 20: are you using `Any`?

flyingsilverfin (approved):

tests/integration/test_grakn.py
lolski: Yes, according to @alexjpwalker , we can't use the word `test` because there's a built-in module with the same name that causes the runtime to be confused. Is that correct Alex?

grakn/query/query_manager.py
lolski: Fixed in https://github.com/graknlabs/client-python/pull/148/commits/6406a2970bd49e9458fd226d93beafafdf37c449

grakn/concept/answer/concept_map.py
lolski: Fixed in https://github.com/graknlabs/client-python/pull/148/commits/6406a2970bd49e9458fd226d93beafafdf37c449

tests/integration/test_grakn.py
alexjpwalker: :))))))))))))

alexjpwalker: Python is fun isn't it

alexjpwalker: Incidentally the reason Ganesh had to make this change is because I forgot to (was planning to delete these tests in the future once any useful parts are migrated)

alexjpwalker (approved):

PR #155, opened 2021-01-11
Update automation.yml to include `behaviour` dependency

VladGan:

Following https://github.com/graknlabs/client-python/pull/154 PR, dependency analysis showed that `behaviour` repo is not defined in dependencies of automation.yml

### DIFF
--- a/.grabl/automation.yml
+++ b/.grabl/automation.yml
@@ -22,6 +22,7 @@ config:
   dependencies:
     dependencies: [build]
     common: [build, release]
+    behaviour: [build]
 
 build:
   quality:


.grabl/automation.yml
```diff
@@ -22,6 +22,7 @@ config:
   dependencies:
     dependencies: [build]
     common: [build, release]
+    behaviour: [build]
```
alexjpwalker, line 25: What does this actually do?

alexjpwalker (approved):

vmax (approved):

PR #161, opened 2021-01-21
Add ability to retrieve all rules

alexjpwalker:

Since Rule was refactored in Grakn 2.0 to no longer be a subtype of Concept, we lost the ability to retrieve all rules using `match $x sub rule; get;`.

So, to allow the user to retrieve all rules, we now add `get_rules` to the `LogicManager`, enabling use of `tx().logic().get_rules()`.

Add get_rules to LogicManager

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -107,7 +107,7 @@ release_validate_python_deps(
     name = "release-validate-python-deps",
     requirements = "//:requirements.txt",
     tagged_deps = [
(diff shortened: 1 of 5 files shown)


vmax (approved):

flyingsilverfin (approved):
just mind the deps, as usual, before we end up releasing

PR #168, opened 2021-01-26
Create one gRPC Channel per session and transaction

alexjpwalker:

We observed that when using 2 parallel threads to open a large number of transactions, response messages from the server would be intermittently dropped, leading to the client stalling. Creating a separate gRPC Channel for each transaction appears to solve the issue.

- Create one gRPC Channel per session
- Create one gRPC Channel per transaction

### DIFF
--- a/grakn/rpc/transaction.py
+++ b/grakn/rpc/transaction.py
@@ -45,14 +45,14 @@ class TransactionType(enum.Enum):
 
 class Transaction:
 
-    def __init__(self, channel: grpc.Channel, session_id: str, transaction_type: TransactionType, options=GraknOptions()):
+    def __init__(self, address: str, session_id: str, transaction_type: TransactionType, options=GraknOptions()):
         self._transaction_type = transaction_type
         self._concept_manager = ConceptManager(self)
         self._query_manager = QueryManager(self)
         self._logic_manager = LogicManager(self)
         self._response_queues = {}
 
-        self._grpc_stub = GraknStub(channel)
+        self._grpc_stub = GraknStub(grpc.insecure_channel(address))
         self._request_iterator = RequestIterator()
         self._response_iterator = self._grpc_stub.transaction(self._request_iterator)
         self._transaction_was_closed = False
--- a/tests/BUILD
+++ b/tests/BUILD
@@ -42,6 +42,17 @@ checkstyle_test(
     size = "small",
 )
 
+py_test(
+    name = "test_concurrent",
+    srcs = [
+        "integration/test_concurrent.py",
+    ],
+    deps = [
+        "//:client_python",
+    ],
+    python_version = "PY3"
+)
+
 artifact_extractor(
     name = "grakn-extractor",
     artifact = ":native-grakn-artifact",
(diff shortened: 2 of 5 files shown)


alexjpwalker:
Note: Do not delete this branch after merging! It is required in order to demonstrate the issue to the gRPC maintainers.

grakn/rpc/transaction.py
```diff
@@ -45,14 +45,14 @@ class TransactionType(enum.Enum):
...
-    def __init__(self, channel: grpc.Channel, session_id: str, transaction_type: TransactionType, options=GraknOptions()):
+    def __init__(self, address: str, session_id: str, transaction_type: TransactionType, options=GraknOptions()):
         self._transaction_type = transaction_type
         self._concept_manager = ConceptManager(self)
         self._query_manager = QueryManager(self)
         self._logic_manager = LogicManager(self)
         self._response_queues = {}
 
-        self._grpc_stub = GraknStub(channel)
+        self._grpc_stub = GraknStub(grpc.insecure_channel(address))
```
alexjpwalker, line 55: This is the main fix here: it seems that using a single gRPC Channel was causing a potential issue where responses from the server would be dropped leading to stalling.

Now, we create a separate Channel for each transaction (and session)

tests/BUILD
```diff
@@ -42,6 +42,17 @@ checkstyle_test(
...
+    name = "test_concurrent",
+    srcs = [
+        "integration/test_concurrent.py",
+    ],
+    deps = [
+        "//:client_python",
+    ],
+    python_version = "PY3"
+)
+
```
alexjpwalker, line 55: This integration test is a reproducible scenario for the error fixed in this PR.

tests/integration/test_concurrent.py
```diff
@@ -0,0 +1,60 @@
...
+from unittest import TestCase
+
+from grakn.client import GraknClient, SessionType, TransactionType
+from grakn.rpc.session import Session
+
+
+class TestConcurrent(TestCase):
+
+    @classmethod
+    def setUpClass(cls):
```
vmax, line 32: @alexjpwalker looks like this method doesn't do much, could we remove it safely?

tests/integration/test_concurrent.py
```diff
@@ -0,0 +1,60 @@
...
+        super(TestConcurrent, cls).setUpClass()
+
+    @classmethod
+    def tearDownClass(cls):
+        super(TestConcurrent, cls).tearDownClass()
```
vmax, line 37: @alexjpwalker looks like this method doesn't do much, could we remove it safely?

tests/integration/test_concurrent.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def setUp(self):
+        pass
```
vmax, line 40: @alexjpwalker looks like this method doesn't do much, could we remove it safely?

tests/integration/test_concurrent.py
```diff
@@ -0,0 +1,60 @@
...
+    def open_tx(self, session: Session, *args):
+        tx = session.transaction(TransactionType.WRITE)
+        tx.close()
+        self.txs_closed += 1
+        print("Total txs closed: %d" % self.txs_closed)
+
+    def test_open_many_transactions_in_parallel(self):
+        self.txs_closed = 0
+        with GraknClient() as client:
+            with client.session("grakn", SessionType.DATA) as session:
```
vmax, line 51: @alexjpwalker in 2.0 unless you create it, it won't exist

vmax (approved):

Issue #172, opened 2021-01-28
read_transaction.concepts().get_thing_type() raises AttributeError: 'ConceptManager' object has no attribute 'get_thing_type'


JonThom:
## Description

When calling `read_transaction.concepts().get_thing_type("relation")` I get 

> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> AttributeError: 'ConceptManager' object has no attribute 'get_thing_type'

## Environment

1. OS (where Grakn server runs): Mac OS 10.14.6
2. Grakn version (and platform): Grakn Core 2.0.0-alpha-3
3. Grakn client-python version: client-python 2.0.0a5
4. Python version: 3.9

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Open a python terminal
2. `from grakn.client import *`
3. `client = GraknClient("127.0.0.1:1729")`
4. `session = client.session(<myschema>, SessionType.SCHEMA)`
5. `tx = session.transaction(TransactionType.READ)`
6. `tx.concepts().get_thing_type("relation")`

## Expected Output

a type object

## Actual Output

> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> AttributeError: 'ConceptManager' object has no attribute 'get_thing_type'

VladGan:
@JonThom thanks for submitting the issue, we fixed it in https://github.com/graknlabs/client-python/pull/174

PR #174, opened 2021-02-01
Make private concept API method public

VladGan:

We fixed the issue with `get_thing_type` API not being available for the user. 

`get_thing_type` method in `ConceptManager` made public.

close https://github.com/graknlabs/client-python/issues/172


grakn/concept/concept_manager.py
```diff
@@ -88,7 +88,7 @@ def get_thing(self, iid: str):
         response = self._execute(req)
         return concept_proto_reader.thing(response.get_thing_res.thing) if response.get_thing_res.WhichOneof("res") == "thing" else None
 
-    def _get_thing_type(self, label: str):
+    def get_thing_type(self, label: str):
```
VladGan, line 91: I assume it was accidental

alexjpwalker: Looks like it!

alexjpwalker (approved):

vmax (approved):

PR #177, opened 2021-02-04
Support new graql update queries

### LINKED ISSUE #176: Enable BDD tests for all Graql Queries on Client Python
## Problem to Solve

Not all Graql query BDD tests are called in CI.

## Proposed Solution

Call all Graql query BDD tests in CI, and make sure they pass.

Phillammon:

We synchronised client-python with the current master versions of protocol, behavioural tests, and grakn core server, to enable support for the new update query type, as well as adding support for related tests and automation.

- Updated protocol version
- Added update queries to querymanager
- Added test steps for rules, relations, and update queries
- Added insert, delete and match to behavioural tests.
- Reformatted automation.yml

Closes #176


.grabl/automation.yml
```diff
@@ -61,10 +61,64 @@ build:
...
+        sudo unlink /usr/bin/python3
+        sudo ln -s $(which python3) /usr/bin/python3
+        sudo ln -s /usr/share/pyshared/lsb_release.py /opt/pyenv/versions/3.6.10/lib/python3.6/site-packages/lsb_release.py
+        export ARTIFACT_USERNAME=$REPO_GRAKN_USERNAME
+        export ARTIFACT_PASSWORD=$REPO_GRAKN_PASSWORD
+        bazel run @graknlabs_dependencies//distribution/artifact:create-netrc
+        bazel test //test/behaviour/graql/language/define:checkstyle --test_output=streamed
+        bazel test //test/behaviour/graql/language/define:test-core --test_output=streamed --jobs=1
+        bazel test //test/behaviour/graql/language/undefine:checkstyle --test_output=streamed
+        bazel test //test/behaviour/graql/language/undefine:test-core --test_output=streamed --jobs=1
```
alexjpwalker: Why are these split up? We can simply write `bazel test //test/behaviour/graql/language/define/...` and the same for undefine. With `--jobs=1` of course. Also there is no need for streamed test output - `--test_output=errors` should be enough.

(For the record, streamed test output results in job logs easily exceeding 10,000 lines)

.grabl/automation.yml
```diff
@@ -80,7 +134,7 @@ build:
         bazel test //tests:test_cluster_failover --test_output=errors
     deploy-pip-snapshot:
       image: graknlabs-ubuntu-20.04
-      dependencies: [build, test-behaviour, test-cluster-failover]
+      dependencies: [build, test-behaviour-connection, test-behaviour-concept, test-behaviour-match, test-behaviour-writable, test-behaviour-definable test-cluster-failover]
```
alexjpwalker, line 137: This is invalid automation.yml right?
```suggestion
      dependencies: [build, test-behaviour-connection, test-behaviour-concept, test-behaviour-match, test-behaviour-writable, test-behaviour-definable, test-cluster-failover]
```

tests/behaviour/background/environment_base.py
```diff
@@ -78,6 +80,9 @@ def after_scenario(context: Context, scenario):
     if scenario.status == Status.skipped:
         return
 
+    #TODO: REMOVE THIS ONCE THE CRASHES ARE FIXED
+    time.sleep(1)
```
alexjpwalker, line 84: `after_scenario` is literally run after every single scenario. Given that each Graql language file has around 60-140 scenarios, and we run Core and Cluster versions of each feature, this will easily add 4 minutes to the overall runtime in CI.

We fixed a similar issue in Cluster by adding a 0.01 second sleep. I would propose we do the same here.
```suggestion
    time.sleep(0.01)
```

tests/behaviour/background/environment_base.py
```diff
@@ -87,4 +92,7 @@ def after_scenario(context: Context, scenario):
 
 
 def after_all(context: Context):
+    #TODO: REMOVE THIS ONCE THE CRASHES ARE FIXED
+    time.sleep(1)
```
alexjpwalker, line 96: See above, but since this is `after_all` it's less important.

alexjpwalker (changes requested):

.grabl/automation.yml
Phillammon: Split up due to cluster tests not presently functioning. Otherwise addressed

tests/behaviour/background/environment_base.py
Phillammon: Modified.

alexjpwalker (approved):

PR #178, opened 2021-02-05
Make DatabaseManager more fault tolerant in Cluster

alexjpwalker:

We've introduced a couple of fault tolerance improvements into the Grakn Cluster version of DatabaseManager:

- `contains` and `all` will now try to retrieve information from every node, instead of failing if the first node is down
- `create` and `delete` will now be applied to every working node, even if some nodes are down.

For now, this means `create` will no longer throw if the database already exists, and `delete` will no longer throw if the database doesn't exist.

- `contains` and `all` will now try to retrieve information from every node, instead of failing if the first node is down
- `create` and `delete` will now be applied to every working node, even if some nodes are down.

### DIFF
--- a/grakn/client.py
+++ b/grakn/client.py
@@ -44,8 +44,8 @@ class GraknClient(ABC):
         return _RPCGraknClient(address)
 
     @staticmethod
-    def cluster(address=DEFAULT_ADDRESS) -> "GraknClient":
-        return _RPCGraknClientCluster(address)
+    def cluster(addresses: List[str]) -> "GraknClient":
+        return _RPCGraknClientCluster(addresses)
 
     @abstractmethod
     def session(self, database: str, session_type: SessionType, options: GraknOptions = None) -> Session:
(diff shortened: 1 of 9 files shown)


tools/behave_rule.bzl
```diff
@@ -64,7 +64,7 @@ def _rule_implementation(ctx):
            echo Starting Grakn Server
            mkdir ./grakn_distribution/"$DIRECTORY"/grakn_test
            ./grakn_distribution/"$DIRECTORY"/grakn server --data grakn_test &
-           sleep 8
+           sleep 9
```
alexjpwalker, line 67: This is a simple CI fix but in a follow-up PR, I will be changing this logic to poll `lsof` until the server has actually started.

grakn/client.py
```diff
@@ -44,8 +44,8 @@ def core(address=DEFAULT_ADDRESS) -> "GraknClient":
         return _RPCGraknClient(address)
 
     @staticmethod
-    def cluster(address=DEFAULT_ADDRESS) -> "GraknClient":
-        return _RPCGraknClientCluster(address)
+    def cluster(addresses: List[str]) -> "GraknClient":
+        return _RPCGraknClientCluster(addresses)
```
alexjpwalker, line 48: This change drops support for constructing `GraknClient.cluster()` with no parameters; but we're fine with that - the default address is pretty nonsensical for Cluster. Developers won't be running Cluster on their machines.

vmax (approved):

PR #180, opened 2021-02-09
When testing Grakn, randomize its port and poll to detect its startup

alexjpwalker:

Previously when running tests, Grakn would always start on port 1729. This rendered it impossible for us to parallelise our tests. Now, we pick a random port number between 40000 and 60000.

Also, previously we waited N milliseconds and then ran the tests. This was wasteful, because often Grakn would start way in advance; and unreliable, because sometimes Grakn would not start in time and CI would fail. Now, we add a polling loop that checks if the server port is in use, and signals that the tests may begin once the port is in use.

- Randomize port when running Grakn during BDD tests
- During BDD tests, poll the ports to detect when Grakn has actually started
- Enable test parallelisation in CI

### DIFF
--- a/.grabl/automation.yml
+++ b/.grabl/automation.yml
@@ -60,7 +60,7 @@ build:
         export ARTIFACT_USERNAME=$REPO_GRAKN_USERNAME
         export ARTIFACT_PASSWORD=$REPO_GRAKN_PASSWORD
         bazel run @graknlabs_dependencies//distribution/artifact:create-netrc
-        bazel test //tests/behaviour/connection/... --test_output=errors
+        bazel test //tests/behaviour/connection/... --test_output=errors --jobs=1
     test-behaviour-concept:
       image: graknlabs-ubuntu-20.04
       type: foreground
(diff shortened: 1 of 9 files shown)


.grabl/automation.yml
```diff
@@ -73,7 +73,7 @@ build:
         export ARTIFACT_USERNAME=$REPO_GRAKN_USERNAME
         export ARTIFACT_PASSWORD=$REPO_GRAKN_PASSWORD
         bazel run @graknlabs_dependencies//distribution/artifact:create-netrc
-        bazel test //tests/behaviour/concept/... --test_output=errors --jobs=1
```
alexjpwalker, line 76: Now that we randomise the port number, we can safely re-enable parallelisation of all tests (except the Connection tests, which intermittently cause Grabl to run out of memory if parallelised)

tools/behave_rule.bzl
```diff
@@ -60,19 +60,57 @@ def _rule_implementation(ctx):
...
+           fi
+
+           echo Successfully unarchived Grakn $PRODUCT distribution.
+
+           RND=20001
+           while [ $RND -gt 20000 ]  # Guarantee fair distribution of random ports
+           do
+             RND=$RANDOM
+           done
+           PORT=$((40000 + $RND))
```
alexjpwalker, line 77: `$RANDOM` returns a random integer in the range `[0..32767]`. This means that when taking `$RANDOM % 20000`, the result is not a fair distribution - for example, `0` is twice as likely as `15000` to be picked.

To mitigate this issue and minimise clashes, we reroll until we get a number in the range `[0..20000]`.

lolski (approved):

PR #181, opened 2021-02-10
Speed up CI

vmax:

Cut the time spent on CI jobs by upgrading to latest `@rules_python` and `pip`

Use latest `pip` (both in standalone tests and in Bazel targets) to ensure that `grpcio` native library is not rebuilt from scratch every time but just fetched from PyPI.

### DIFF
--- a/.grabl/automation.yml
+++ b/.grabl/automation.yml
@@ -53,6 +53,7 @@ build:
       type: foreground
       command: |
         pyenv global 3.6.10
+        pip3 install -U pip
         pip install -r requirements_dev.txt
         sudo unlink /usr/bin/python3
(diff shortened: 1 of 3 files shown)


.grabl/automation.yml
```diff
@@ -53,6 +53,7 @@ build:
       type: foreground
       command: |
         pyenv global 3.6.10
+        pip3 install -U pip
```
alexjpwalker, line 56: Why are we upgrading `pip` in each run (as opposed to upgrading pip in our image)? (Or have we actually done both)

vmax: @alexjpwalker I'll this to be a separate step in the image generation and then we can remove it

alexjpwalker (approved):

PR #183, opened 2021-02-11
Bind Cluster sessions to single server nodes, and add naive load balancing

alexjpwalker:

Previously, each Cluster session would lazily create Core sessions on-demand as Transactions were opened. This delegated too much responsibility to the Transaction. Now, each Cluster session is bound to a single server node, which may be a primary or a secondary node depending on the Options passed to the Session.

We also select secondary nodes in a fairer, more balanced way; the server indicates which node is its first preference for secondary sessions, and the client tries to obey the server's decision, with fallback to backup nodes only in the event of a failure.

- On creating a Cluster session, a Core session is immediately created to a server node of the appropriate type
- Node selection is based on the `read_any_replica` Option passed to the Session
- When `read_any_replica` is `true`, the client will first select nodes that the server has indicated are "preferred secondary", with fallback to backup nodes only in the event of a failure

### DIFF
--- a/grakn/client.py
+++ b/grakn/client.py
@@ -27,10 +27,12 @@ from grpc import RpcError
 # Repackaging these symbols allows them to be imported from "grakn.client"
 from grakn.common.exception import GraknClientException  # noqa # pylint: disable=unused-import
 from grakn.concept.type.value_type import ValueType  # noqa # pylint: disable=unused-import
-from grakn.options import GraknOptions
+from grakn.options import GraknOptions, GraknClusterOptions
+from grakn.rpc.cluster.failsafe_task import FailsafeTask
+from grakn.rpc.cluster.replica_info import ReplicaInfo
 from grakn.rpc.cluster.server_address import ServerAddress
 from grakn.rpc.cluster.database_manager import _DatabaseManagerClusterRPC
-from grakn.rpc.cluster.session import _SessionClusterRPC
+from grakn.rpc.cluster.session import SessionClusterRPC
 from grakn.rpc.database_manager import DatabaseManager, _DatabaseManagerRPC
 from grakn.rpc.session import Session, SessionType, _SessionRPC
 from grakn.rpc.transaction import TransactionType  # noqa # pylint: disable=unused-import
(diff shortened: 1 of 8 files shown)


grakn/client.py
```diff
@@ -149,16 +161,27 @@ def core_client(self, address: ServerAddress) -> _ClientRPC:
...
+
+class _OpenSessionFailsafeTask(FailsafeTask):
+
+    def __init__(self, database: str, session_type: SessionType, options: GraknClusterOptions, client: "_ClientClusterRPC"):
+        super().__init__(client, database)
+        self.session_type = session_type
+        self.options = options
+
+    def run(self, replica: ReplicaInfo.Replica):
+        return SessionClusterRPC(self.client, replica.address(), self.database, self.session_type, self.options)
```
alexjpwalker, line 187: `FailsafeTask` is a new class that holds the failover/retry logic. It is an abstract base class, requiring that `run` is implemented, where `run` represents a unit of work that can fail and should be retried / failed over.

grakn/rpc/cluster/failsafe_task.py
```diff
@@ -0,0 +1,123 @@
...
+            if retries > self.PRIMARY_REPLICA_TASK_MAX_RETRIES:
+                raise self._cluster_not_available_exception()
+
+    def run_any_replica(self):
+        if self.database in self.client.replica_info_map():
+            replica_info = self.client.replica_info_map()[self.database]
+        else:
+            replica_info = self._fetch_database_replicas()
+
+        replicas = [replica_info.preferred_secondary_replica()] + [replica for replica in replica_info.replicas() if not replica.is_preferred_secondary()]
```
alexjpwalker, line 82: When performing an operation in an "any replica" session, we now give the _preferred secondary replica_ priority over other replicas. It will be contacted first, and the other replicas will only be contacted in the event of a failure.

grakn/rpc/cluster/session.py
```diff
@@ -16,137 +16,68 @@
...
+
+    def run(self, replica: ReplicaInfo.Replica):
+        return self.cluster_session.core_session.transaction(self.transaction_type, self.options)
+
+    def rerun(self, replica: ReplicaInfo.Replica):
+        if self.cluster_session.core_session:
+            self.cluster_session.core_session.close()
+        self.cluster_session.core_client = self.cluster_session.cluster_client.core_client(replica.address())
+        self.cluster_session.core_session = self.cluster_session.core_client.session(self.database, self.cluster_session.session_type(), self.options)
+        return self.cluster_session.core_session.transaction(self.transaction_type, self.options)
```
alexjpwalker, line 83: In `FailsafeTask`, `run` is called on the first run, and `rerun` is called on every subsequent run, if the first run fails.

In this case, if the first run fails, we need to tell the Cluster session to:

1) close its Core session
2) switch to a different Core client (linked to a different server node)
3) try to open a new Core session in the new client
4) try to open the transaction using this Core session, instead of the original one.

This allows us to fail-over the opening of Transactions, while enforcing that each Cluster session has, at any one time, precisely one Core session associated with it.

grakn/rpc/cluster/failsafe_task.py
vmax: @alexjpwalker we could maybe simplify this a bit by sorting based on `is_preferred_secondary()` boolean value; something like this:

```suggestion
        replicas = sorted(replica_info.replicas(), key=lambda replica: not replica.is_preferred_secondary()]
```

This would put all replicas with `is_preferred_secondary` first in the list

grakn/rpc/cluster/replica_info.py
```diff
@@ -41,22 +41,30 @@ def primary_replica(self) -> Optional["ReplicaInfo.Replica"]:
         primaries = [replica for replica in self._replicas.values() if replica.is_primary()]
         return max(primaries, key=lambda r: r.term) if primaries else None
 
+    def preferred_secondary_replica(self) -> "ReplicaInfo.Replica":
+        return next(iter([replica for replica in self._replicas.values() if replica.is_preferred_secondary()]), next(iter(self._replicas.values())))
```
vmax, line 45: @alexjpwalker type annotation indicates you're returning one thing but it looks like you're returning a tuple. Could you elaborate on what's supposed to be returned?

grakn/rpc/cluster/failsafe_task.py
alexjpwalker: I think I prefer the current approach, for two reasons:

- It emphasises that there is precisely one preferred secondary replica
- It's clear - we don't need to think about what the sort order is

grakn/rpc/cluster/replica_info.py
alexjpwalker: It's not a tuple, it's: "take the first replica that is a preferred secondary, if it exists; else, just take the first replica"

grakn/rpc/cluster/failsafe_task.py
vmax: @alexjpwalker ok I can agree with the argument. It would be more clear by being more verbose.

grakn/rpc/cluster/replica_info.py
vmax: @alexjpwalker 

>It's not a tuple

So you're saying it returns just one instance?

>take the first replica that is a preferred secondary, if it exists

But if it doesn't exist we'll do `next(iter([]))` which will raise `StopIteration`

I think we could rewrite it for clarity too:

```
for replica in self._replicas.values():
    if replica.is_preferred_secondary():
        return replica
return next(iter(self._replicas.values())))
```

alexjpwalker: > But if it doesn't exist we'll do next(iter([])) which will raise StopIteration

This isn't true - if you do:

`next([], 1)`

it returns `1`, it doesn't throw.

alexjpwalker: And yes, it's just a `next` call so it always returns a single instance.

vmax: @alexjpwalker got it! it's a **single** `next` call, second argument to which is **also** a `next` call. What a Python riddle!

vmax (approved):

PR #186, opened 2021-02-17
Expose replica info when getting Cluster databases

alexjpwalker:

We now expose additional information when fetching databases in Cluster. The database name, server address, primary/secondary state, and term number are all available.

This is a **breaking change**. `client.databases().all()` now returns `List[Database]` (previously it was `List[str]`). To get the old functionality (listing the names), use `[db.name() for db in client.databases().all()]`.

Additionally, `client.databases().delete()` has been deleted, and superseded by the new syntax: `client.databases().get(name).delete()`.

- Delete `client.databases().delete`, and add `client.databases().get`, which returns a new `Database` object that has a `delete` method
- Add `GraknClientCluster` and `DatabaseManagerCluster`, which return `DatabaseCluster` objects when `all` and `get` are called. `DatabaseCluster` objects expose their database name, server address, primary/secondary state and term number.


grakn/rpc/cluster/database.py
```diff
@@ -0,0 +1,134 @@
+#
```
alexjpwalker, line 1: Renamed from `replica_info`

grakn/rpc/database.py
```diff
@@ -0,0 +1,91 @@
...
+        pass
+
+    class Replica(ABC):
+
+        @abstractmethod
+        def database(self) -> "DatabaseCluster":
+            pass
+
+        @abstractmethod
+        def term(self) -> int:
```
alexjpwalker, line 45: Previously this data was hidden from the user; now it is exposed, for troubleshooting purposes.

grakn/rpc/database_manager.py
```diff
@@ -37,11 +39,11 @@ def create(self, name: str) -> None:
         pass
 
     @abstractmethod
-    def delete(self, name: str) -> None:
```
alexjpwalker, line 40: `delete` no longer exists in `DatabaseManager`. It is now a method on `Database`.

grakn/rpc/database_manager.py
```diff
@@ -37,11 +39,11 @@ def create(self, name: str) -> None:
...
+    def get(self, name: str) -> Database:
```
alexjpwalker, line 42: In Core, the primary purpose of `get` is to retrieve a Database so that you can `delete` it. In Cluster, it is quite a bit more interesting.

requirements.txt
```diff
@@ -26,6 +29,7 @@
 
 ## Dependencies
 
-grakn-protocol==2.0.0a9
+# IMPORTANT: Any changes to these dependencies should be copied to requirements_dev.txt.
```
alexjpwalker, line 32: People kept forgetting to update `requirements_dev.txt`, so now I've left a glaring message telling them to do so.

tests/integration/test_cluster_failover.py
```diff
@@ -21,44 +21,36 @@
...
+    def get_primary_replica(self, database_manager: DatabaseManagerCluster):
+        retry_num = 0
+        while retry_num < 10:
             print("Discovering replicas for database 'grakn'...")
-            res = cluster_grpc_stub.database_replicas(db_replicas_req)
-            dbs = ReplicaInfo.of_proto(res)
-            print("Discovered " + str([str(replica) for replica in dbs.replicas()]))
-            primary_replica = next(iter([replica for replica in dbs.replicas() if replica.is_primary()]), None)
-            if primary_replica:
-                return primary_replica
```
alexjpwalker, line 52: As a consequence of the refactor, it's now much easier for users to get information like which Cluster node holds the current primary replica for a particular database.

alexjpwalker (commented):
These changes look good

vmax (approved):

PR #194, opened 2021-02-26
Fix possible ValueError when closing a session

alexjpwalker:

We patched a flaw where very occasionally, `Session.close` would throw a `ValueError` if a pulse was in flight.

- If cancelling the pulse task throws an error in Session.close, swallow it

### DIFF
--- a/grakn/rpc/session.py
+++ b/grakn/rpc/session.py
@@ -117,8 +117,10 @@ class _SessionRPC(Session):
             self._is_open = False
             self._client.remove_session(self)
             self._lock.release()
-            self._scheduler.cancel(self._pulse)
-            self._scheduler.empty()
+            try:
+                self._scheduler.cancel(self._pulse)
+            except ValueError:  # This may occur if a pulse is in transit but not responded to yet.
+                pass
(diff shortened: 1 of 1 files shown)


grakn/rpc/session.py
```diff
@@ -117,8 +117,10 @@ def close(self) -> None:
             self._is_open = False
             self._client.remove_session(self)
             self._lock.release()
-            self._scheduler.cancel(self._pulse)
-            self._scheduler.empty()
+            try:
+                self._scheduler.cancel(self._pulse)
+            except ValueError:  # This may occur if a pulse is in transit but not responded to yet.
+                pass
```
alexjpwalker, line 123: Given that scheduled tasks run on a separate thread, we need to cancel the task in a thread-safe way. This means that the following (pseudo-) code would not be safe:
```py
if self._pulse in self._scheduler._queue:
    self._scheduler.cancel(self._pulse)
```
Using `try-except` allows us to atomically dequeue the task if it is in the queue, and perform no operation otherwise.

vmax (approved):

PR #200, opened 2021-03-05
Add 'get_type' to local Thing concepts

alexjpwalker:

Previously, `get_type` was only available on remote `Thing` concepts. Now, we include type information whenever a Graql query or Concept API method returns a `Thing`, enabling users to use it without an additional network roundtrip per concept.

Add 'get_type' to local Thing concepts

### DIFF
--- a/grakn/concept/thing/attribute.py
+++ b/grakn/concept/thing/attribute.py
@@ -16,7 +16,6 @@
 # specific language governing permissions and limitations
 # under the License.
 #
-
 from datetime import datetime
 
 import grakn_protocol.protobuf.concept_pb2 as concept_proto
@@ -26,6 +25,9 @@ from grakn.concept.thing.thing import Thing, RemoteThing
 
 
 class Attribute(Thing):
+
+    def get_value_type(self):
+        return self.get_type().get_value_type()
 
     def is_attribute(self):
         return True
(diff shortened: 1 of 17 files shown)


grakn/concept/thing/attribute.py
```diff
@@ -27,6 +26,9 @@
 
 class Attribute(Thing):
 
+    def get_value_type(self):
+        return self.get_type().get_value_type()
```
flyingsilverfin, line 30: hmmm i don't think we should expose this here - the users themselves can get the value type via the type. we're generally not providing shorthands unless it has overwhemling benefits/usage.

grakn/concept/thing/attribute.py
```diff
@@ -48,6 +50,9 @@ def is_datetime(self):
 
 class RemoteAttribute(RemoteThing):
 
+    def get_value_type(self):
+        return self.get_type().get_value_type()
+
```
flyingsilverfin, line 55: same here, the users can go the type themselves

flyingsilverfin (changes requested):
left a comment on shorthand methods

lolski (approved):

flyingsilverfin (approved):

PR #211, opened 2021-03-26
Set default parallelism of CoreClient to 2

alexjpwalker:

Given that Python doesn't provide an easy way of getting the number of CPUs available to the process, we decided to reduce the default parallelism of CoreClient from 3 to 2. This should ensure it continues to run smoothly on systems with fewer cores.

Set default parallelism of CoreClient to 2


grakn/core/client.py
```diff
@@ -30,8 +30,8 @@
 
 
 def calculate_parallelisation():
-    # TODO: cores = Runtime.getRuntime().availableProcessors()
-    return 3
+    # TODO: Detect number of available CPUs
+    return 2
```
alexjpwalker, line 34: Given that Python doesn't provide an easy way of getting the number of CPUs available to the process, we decided to reduce the default parallelism of CoreClient from 3 to 2. This should ensure it continues to run smoothly on systems with fewer cores.

flyingsilverfin (approved):

PR #212, opened 2021-03-29
Shorten enum names to SessionType, TransactionType

alexjpwalker:

While aligning Client Python with Client Java, we introduced `GraknTransaction.Type` and `GraknSession.Type`. But in Python these proved cumbersome to type out. So we reverted them to `SessionType` and `TransactionType`.

Shorten enum names to SessionType, TransactionType


flyingsilverfin (approved):

alexjpwalker (commented):
These changes look good

PR #225, opened 2021-06-04
Implement cluster authentication

flyingsilverfin:

Implemented authentication secure channels, and user management for typedb cluster.
**SSL is now required for connections to TypeDB cluster** , because sending username/passwords over plaintext connections is disallowed by gRPC.

* Refactor packages so that session, transaction, database etc. now live under `//api/connection`
* match the package structure by creating `//connection/cluster` and `//connection/core`
* extract shared implementation into `//connection/*.py`
* create API and implementations for `User` and `UserManager`
* Implement secure channel and authentication mechanisms for cluster
* implement BDD steps for user management
* bump gRPC to `1.38.0`


tests/integration/test_cluster_failover.py
```diff
@@ -35,7 +36,9 @@
 class TestClusterFailover(TestCase):
 
     def setUp(self):
-        with TypeDB.cluster_client(["localhost:11729", "localhost:21729", "localhost:31729"]) as client:
+        root_ca_path = os.environ["ROOT_CA"]
+        credential = TypeDBCredential("admin", "password", root_ca_path)
+        with TypeDB.cluster_client(["127.0.0.1:11729", "127.0.0.1:21729", "127.0.0.1:31729"], credential) as client:
```
flyingsilverfin, line 41: note to future self: must use IP to match built-in server credential, not `localhost`

tools/cluster_test_rule.bzl
```diff
@@ -88,7 +91,10 @@ def _rule_implementation(ctx):
     cmd += """
            echo Tests concluded with exit value $RESULT
            echo Stopping servers.
-           kill $(jps | awk '/TypeDBNode/ {print $1}' | paste -sd " " -)
+           procs=$(jps | awk '/TypeDBNode/ {print $1}' | paste -sd " " -)
+           if [ -n "$procs" ]; then
+             kill $procs
+           fi
```
flyingsilverfin, line 97: note to future self: we needed to add this because the cluster failover test can kill the servers started by this script and restart them from within the test. This means by the time we get to `kill` all the original TypeDB processes will already be dead and the `kill` command will fail because there are no processes left

PR #229, opened 2021-07-15
Add concept cast methods, and add all Concept types to typedb.client

alexjpwalker:

We added methods to cast concept types to each other: for example `concept.as_attribute()` will now cast a `Concept` to an `Attribute`, throwing an exception if it is not actually an `Attribute`. This aids type checking systems.

Also, all the `Concept` types are now importable from `typedb.client` directly.

Add concept cast methods, and add all Concept types to typedb.client

### DIFF
--- a/typedb/api/concept/thing/attribute.py
+++ b/typedb/api/concept/thing/attribute.py
@@ -20,7 +20,7 @@
 #
 from abc import ABC, abstractmethod
 from datetime import datetime
-from typing import TYPE_CHECKING, Iterator
+from typing import TYPE_CHECKING, Iterator, Union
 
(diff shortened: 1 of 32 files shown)


typedb/api/concept/thing/attribute.py
```diff
@@ -37,6 +37,10 @@ class Attribute(Thing, ABC):
     def get_type(self) -> "AttributeType":
         pass
 
+    @abstractmethod
+    def get_value(self) -> Union[bool, int, float, str, datetime]:
```
flyingsilverfin, line 41: gud

flyingsilverfin (approved):

PR #231, opened 2021-08-19
Fix incorrect usages of TypeDBClientException

alexjpwalker:

We fixed incorrect usages of TypeDBClientException leading to unhelpful error messages in some scenarios.

Fix incorrect usages of TypeDBClientException leading to unhelpful error messages in some scenarios.

### DIFF
--- a/typedb/common/exception.py
+++ b/typedb/common/exception.py
@@ -25,7 +25,7 @@ from grpc import RpcError, Call, StatusCode
 
 class TypeDBClientException(Exception):
 
-    def __init__(self, msg: Union["ErrorMessage", str], cause: Optional[BaseException], params: Any = None):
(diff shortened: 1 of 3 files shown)


typedb/common/exception.py
```diff
@@ -25,7 +25,7 @@
 
 class TypeDBClientException(Exception):
 
-    def __init__(self, msg: Union["ErrorMessage", str], cause: Optional[BaseException], params: Any = None):
+    def __init__(self, msg: Union["ErrorMessage", str], cause: BaseException = None, params: Any = None):
```
alexjpwalker, line 28: Given that Python is so easy to misuse, we may as well make the class a little more resilient against misuse.

vmax (approved):

PR #233, opened 2021-09-29
Remove connection factory classes

lolski:

We've renamed the terminology 'name' to 'username' since it's more appropriate.

We've also simplified the architecture of both Core and ClusterServer clients by moving the `channel` and `stub` creation logic from the connection factory classes directly into the client classes.

- Renaming the terminology 'name' to 'username' since it's more appropriate: the terminology 'name' is typically used to refer to the user's real / birth name
- We've also simplified the architecture of both Core and ClusterServer clients by moving the `channel` and `stub` creation logic from the connection factory classes directly into the client classes. Before this PR, the gRPC `channel` and `stub` of a TypeDB client is constructed from a "connection factory" instance. A `_CoreClient` instance for example, requires a `_CoreConnectionFactory` instance. Similarly, a `_ClusterServerClient` depends on a `_ClusterConnectionFactory`. Getting rid of the "connection factories" makes the architecture much simpler.


typedb/connection/cluster/client.py
```diff
@@ -41,20 +40,17 @@ class _ClusterClient(TypeDBClusterClient):
     def __init__(self, addresses: Iterable[str], credential: TypeDBCredential, parallelisation: int = None):
         self._credential = credential
         self._server_clients: Dict[str, _ClusterServerClient] = {addr: _ClusterServerClient(addr, credential, parallelisation) for addr in self._fetch_server_addresses(addresses)}
-        self._stubs = {addr: client.connection_factory().newTypeDBStub(client.channel()) for (addr, client) in self._server_clients.items()}
```
lolski, line 44: We're no longer keeping the stubs list. It is redundant given that we're already keeping a list of clients (each of which has a stub).

typedb/connection/cluster/client.py
```diff
@@ -41,20 +40,17 @@ class _ClusterClient(TypeDBClusterClient):
...
         self._database_managers = _ClusterDatabaseManager(self)
         self._cluster_databases: Dict[str, _ClusterDatabase] = {}
         self._user_manager = _ClusterUserManager(self)
         self._is_open = True
-        print("Cluster client created")
```
lolski, line 49: Removing this debug print

typedb/connection/cluster/client.py
```diff
@@ -41,20 +40,17 @@ class _ClusterClient(TypeDBClusterClient):
...
 
     def _fetch_server_addresses(self, addresses: Iterable[str]) -> Set[str]:
         for address in addresses:
             try:
                 print("Fetching list of cluster servers from %s..." % address)
                 with _ClusterServerClient(address, self._credential) as client:
-                    typedb_cluster_stub = client.connection_factory().newTypeDBStub(client.channel())
-                    res = typedb_cluster_stub.servers_all(cluster_server_manager_all_req())
+                    res = client.stub().servers_all(cluster_server_manager_all_req())
```
lolski, line 53: Reusing the same `stub()` from the `client` object, rather than creating a new one.

typedb/connection/transaction.py
```diff
@@ -50,10 +50,7 @@ def __init__(self, session: "_TypeDBSessionImpl", transaction_type: TransactionT
         self._logic_manager = _LogicManager(self)
 
         try:
-            # Other TypeDBClient implementations reuse a single gRPC Channel, but the Python client stalls
-            # when opening several transactions in parallel from one Channel.
-            stub = session.client().connection_factory().newTypeDBStub(session.client().connection_factory().newChannel(session.address()))
-            self._bidirectional_stream = BidirectionalStream(stub, session.transmitter())
+            self._bidirectional_stream = BidirectionalStream(session.client().stub(), session.transmitter())
```
lolski, line 53: Reusing an existing stub rather than creating a new one.

You wrote a comment above the line saying that doing so may cause the client to stall, but the test seem to pass without stalling. Therefore it seems that my change is working, right @alexjpwalker ?

alexjpwalker: Hmm, actually, I don't think this bug is covered in BDD.

It's covered in an integration test, which I stupidly seem to have forgotten to merge into master! See: https://github.com/alexjpwalker/typedb-client-python/blob/stalling-fix/tests/integration/test_concurrent.py

It opens 10000 transactions in parallel, closing each one -at some point- after opening.

Let's reintroduce the test and try it out locally?

alexjpwalker (changes requested):

alexjpwalker (approved):

PR #235, opened 2021-09-30
Enable faster authentication with "user token" mechanism

lolski:

Authenticating a user by verifying their password introduces a significant overhead, which comes from the fact that the password is hashed with cryptographic hash function. In effect, opening a new session or transaction becomes much slower.

We've now improved the speed of user-authentication by introducing a mechanism for verifying user credential with cheap "user token" rather than password.

- `ClusterServerStub` performs an attempt to retrieve user token when instantiated. Should the attempt fail, it will be retried the next time.
- The `resilient_call` method of `ClusterServerStub` may renew the token and then reattempt the call if the token is no longer valid

### DIFF
--- a/typedb/common/rpc/stub.py
+++ b/typedb/common/rpc/stub.py
@@ -20,14 +20,16 @@
 #
 
 from abc import ABC
-from typing import Callable, Iterator, TypeVar
+from typing import Iterator
+from typing import TypeVar, Callable
 
 import typedb_protocol.common.session_pb2 as session_proto
 import typedb_protocol.common.transaction_pb2 as transaction_proto
 import typedb_protocol.core.core_database_pb2 as core_database_proto
 import typedb_protocol.core.core_service_pb2_grpc as core_service_proto
-from grpc import Channel
+from grpc import Channel, RpcError
 
+from typedb.common.exception import TypeDBClientException
 
 T = TypeVar('T')
 
@@ -68,4 +70,8 @@ class TypeDBStub(ABC):
         pass
 
     def resilient_call(self, function: Callable[[], T]) -> T:
-        pass
+        try:
+            # TODO actually implement forced gRPC to reconnected rapidly, which provides resilience
+            return function()
+        except RpcError as e:
+            raise TypeDBClientException.of_rpc(e)
--- a/typedb/common/rpc/stub.py
+++ b/typedb/common/rpc/stub.py
@@ -20,60 +20,49 @@
 #
 
 from abc import ABC
-from typing import TypeVar, Callable, Iterator
-
+from typing import Callable, Iterator
 
 import typedb_protocol.common.session_pb2 as session_proto
 import typedb_protocol.common.transaction_pb2 as transaction_proto
 import typedb_protocol.core.core_database_pb2 as core_database_proto
 import typedb_protocol.core.core_service_pb2_grpc as core_service_proto
-
-from grpc import Channel, RpcError
-
-from typedb.common.exception import TypeDBClientException
-
-T = TypeVar('T')
-
-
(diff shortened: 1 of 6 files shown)


typedb/common/rpc/stub.py
```diff
@@ -20,58 +20,52 @@
...
+from grpc import Channel
 
-from grpc import Channel, RpcError
-
-from typedb.common.exception import TypeDBClientException
 
 T = TypeVar('T')
 
 
-def resilient_call(function: Callable[[], T]) -> T:
```
lolski, line 38: `resilient_call` is now made to be an abstract method of `TypeDBStub`. `CoreStub` and `ClusterServerStub` will implement them in a different way.

typedb/connection/cluster/server_client.py
```diff
@@ -56,7 +56,7 @@ def _new_channel(self) -> grpc.Channel:
             channel_credentials = grpc.ssl_channel_credentials()
         combined_credentials = grpc.composite_channel_credentials(
             channel_credentials,
-            grpc.metadata_call_credentials(_CredentialAuth(self._credential.username(), self._credential.password()))
+            grpc.metadata_call_credentials(_CredentialAuth(lambda: self._credential, lambda: self._stub.token()))
```
lolski, line 59: The authentication metadata will either contain the credential (eg., username / password), or token.

lolski: In practice, the credential is only sent during the `user_token_renew` call. Once the `token` has been renewed, it will be preferred.

typedb/connection/cluster/server_client.py
```diff
@@ -66,9 +66,13 @@ def close(self) -> None:
 
 
 class _CredentialAuth(grpc.AuthMetadataPlugin):
-    def __init__(self, username, password):
-        self._username = username
-        self._password = password
+    def __init__(self, credential, token):
```
alexjpwalker, line 69: I know this is gRPC's fault and not ours, but how about we diverge from our established principle of following gRPC's naming in this case? The fact that the parameter named `credential` is in fact a `Callable` is... very off-putting.
```suggestion
    def __init__(self, credential_fn, token_fn):
```

alexjpwalker: (Also, if you happen to know the correct type hints, that would be a bonus, but it's no big deal if you don't!)

typedb/connection/cluster/stub.py
```diff
@@ -18,51 +18,84 @@
...
+class _ClusterServerStub(TypeDBStub):
 
-    @staticmethod
-    def create(channel: Channel):
-        return _ClusterServerStub(
-            channel,
-            core_service_proto.TypeDBStub(channel),
-            cluster_service_proto.TypeDBClusterStub(channel)
-        )
+    def __init__(self, channel: Channel, credential: TypeDBCredential):
```
alexjpwalker, line 39: This is a great change!

typedb/connection/cluster/stub.py
```diff
@@ -18,51 +18,84 @@
...
+    def channel(self) -> Channel:
+        return self._channel
+
+    def stub(self) -> TypeDBStub:
+        return self._stub
+
+    def token(self):
+        return self._token
 
+    def resilient_call(self, function: Callable[[], T]) -> T:
```
alexjpwalker, line 86: The logic seems sound, but I'm not convinced that this method is the right place to put that logic.

`resilient_call` is so named because in `client-java`, we ensure that the connection is re-established if it has dropped. Checking authentication seems a little too far from that.

alexjpwalker (changes requested):

alexjpwalker (approved):

PR #244, opened 2021-12-21
Implement and test transaction timeouts

flyingsilverfin:

We implement options for transaction timeout configuration (https://github.com/vaticle/typedb/issues/6487). We also add BDD implementation for new steps required to test transaction timeout and configuring transaction options.

* Add option for transaction timeout (default: 5 mins)
* drain and throw exceptions when the transaction is closed with un-received errors
* add missing BDD steps for session and transaction option configuration


dependencies/vaticle/artifacts.bzl
```diff
@@ -29,7 +29,7 @@ def vaticle_typedb_artifacts():
         artifact_name = "typedb-server-{platform}-{version}.{ext}",
         tag_source = deployment["artifact.release"],
         commit_source = deployment["artifact.snapshot"],
-        commit = "6ed020e52fe379d1100f64511805ed344c7a68db",
+        commit = "2367157bdd898e474198726d0ef5446372a73314",
```
flyingsilverfin, line 32: new typedb that supports transaction timeouts

requirements.txt
```diff
@@ -24,13 +24,14 @@
...
 # Allow importing of snapshots
 --extra-index-url https://repo.vaticle.com/repository/pypi-snapshot/simple
 
 
 ## Dependencies
 
 # IMPORTANT: Any changes to these dependencies should be copied to requirements_dev.txt.
-typedb-protocol==2.5.0
-grpcio==1.38.0
+typedb-protocol==0.0.0-4ce4e1339995c8982914e9fd3177273b693609b7
```
flyingsilverfin, line 35: new protocol with transaction timeouts

tests/behaviour/background/cluster/environment.py
```diff
@@ -42,6 +42,9 @@ def before_scenario(context: Context, scenario):
             scenario.skip("tagged with @" + tag)
             return
     environment_base.before_scenario(context, scenario)
+    context.session_options = TypeDBOptions.cluster().set_infer(True)
+    context.transaction_options = TypeDBOptions.cluster().set_infer(True)
```
flyingsilverfin, line 46: we now have session and transaction options we have configure before each scenario

tests/behaviour/background/environment_base.py
```diff
@@ -46,6 +46,10 @@ def before_scenario(context: Context, scenario):
     context.put = lambda var, thing: _put_impl(context, var, thing)
     context.get_thing_type = lambda root_label, type_label: _get_thing_type_impl(context, root_label, type_label)
     context.clear_answers = lambda: _clear_answers_impl(context)
+    context.option_setters = {
+        "session-idle-timeout-millis": lambda option, value: option.set_session_idle_timeout_millis(int(value)),
+        "transaction-timeout-millis": lambda option, value: option.set_transaction_timeout_millis(int(value)),
+    }
```
flyingsilverfin, line 52: lambdas to set option values in the option objects

tests/behaviour/connection/session/session_steps.py
```diff
@@ -147,3 +147,14 @@ def step_impl(context: Context):
...
+
+######################################
+# session configuration              #
+######################################
+
+@step("set session option {option} to: {value}")
+def step_impl(context: Context, option: str, value: str):
+    if option not in context.option_setters:
+        raise Exception("Unrecognised option: " + option)
+    context.option_setters[option](context.session_options, value)
```
flyingsilverfin, line 159: new BDD testing steps to configure session and transaction options

tests/behaviour/context.py
```diff
@@ -51,12 +51,15 @@ def __init__(self):
         self.sessions_to_transactions: Dict[TypeDBSession, List[TypeDBTransaction]] = {}
         self.sessions_parallel: List[Future[TypeDBSession]] = []
         self.sessions_parallel_to_transactions_parallel: Dict[Future[TypeDBSession], List[TypeDBTransaction]] = {}
+        self.session_options: Optional[TypeDBOptions] = None
+        self.transaction_options: Optional[TypeDBOptions] = None
```
flyingsilverfin, line 55: these have to be set to None and _RE SET_ before each test, otherwise cucumber somehow doesn't propagate these values across scenarios

tools/behave_rule.bzl
```diff
@@ -81,7 +81,7 @@ def _rule_implementation(ctx):
            echo Starting TypeDB $PRODUCT Server.
            mkdir ./typedb_distribution/"$DIRECTORY"/typedb_test
            if [[ $PRODUCT == "Core" ]]; then
-             ./typedb_distribution/"$DIRECTORY"/typedb server --port $PORT --data typedb_test &
+             ./typedb_distribution/"$DIRECTORY"/typedb server --server.address 0.0.0.0:$PORT --storage.data typedb_test &
```
flyingsilverfin, line 84: fix the typedb runner used by behaviour tests with new cli args

typedb/api/connection/options.py
```diff
@@ -46,6 +47,42 @@ def cluster() -> "TypeDBClusterOptions":
...
+        self.session_idle_timeout_millis = session_idle_timeout_millis
+        return self
+
+    def set_transaction_timeout_millis(self, transaction_timeout_millis: int):
+        self.transaction_timeout_millis = transaction_timeout_millis
+        return self
+
+    def set_schema_lock_acquire_timeout_millis(self, schema_lock_acquire_timeout_millis: int):
+        self.schema_lock_acquire_timeout_millis = schema_lock_acquire_timeout_millis
+        return self
```
flyingsilverfin: implement fluent interface like in java to configure options, mostly to allow simple lambdas to configure options in BDD, but also a useful API to have

flyingsilverfin: since we have setters, we also add getters

flyingsilverfin:
Will be post-reviewed by @lolski  or @alexjpwalker later

.grabl/automation.yml
```diff
@@ -64,20 +64,20 @@ build:
...
-        pip3 install -U pip
-        pip install -r requirements_dev.txt
-        sudo unlink /usr/bin/python3
-        sudo ln -s $(which python3) /usr/bin/python3
-        sudo ln -s /usr/share/pyshared/lsb_release.py /opt/pyenv/versions/3.6.10/lib/python3.6/site-packages/lsb_release.py
-        export ARTIFACT_USERNAME=$REPO_VATICLE_USERNAME
-        export ARTIFACT_PASSWORD=$REPO_VATICLE_PASSWORD
-        bazel run @vaticle_dependencies//distribution/artifact:create-netrc
-        .grabl/test-cluster.sh //tests/behaviour/connection/... --test_output=errors --jobs=1
+#    test-behaviour-connection-cluster:
```
alexjpwalker, line 67: Why did we comment these out? Is it just because they are failing? If so, we should keep them enabled and accept that CI jobs fail - it's more dangerous to leave them commented and forget about them

(on the other hand if CI requires them to pass in order to release, then we should temporarily disable them, and immediately bring them back in a follow-up commit)

tests/behaviour/connection/connection_steps.py
```diff
@@ -32,3 +33,8 @@ def step_impl(context: Context):
 @step("connection does not have any database")
 def step_impl(context: Context):
     assert len(context.client.databases().all()) == 0
+
+
+@step("wait {seconds} seconds")
+def step_impl(context: Context, seconds: str):
```
alexjpwalker, line 39: We should create `util_steps.py` and move this step there, as we did in Client Java

typedb/api/connection/options.py
```diff
@@ -46,6 +47,69 @@ def cluster() -> "TypeDBClusterOptions":
     def is_cluster(self) -> bool:
         return False
 
+    def get_infer(self) -> Optional[bool]:
```
alexjpwalker, line 50: The reason I didn't add these methods in the first place was because accessing the attributes (akin to fields in Java) directly is more "Pythonic".

I would recommend deleting them, but we may want a broader discussion on it.

The main thing missing from the methods (which is still missing in this implementation!) is value validation. But this can be achieved in Python using `@property` and `@batch_size.setter` annotations (which I didn't know about / wasn't comfortable using when I originally wrote `options.py`)

typedb/stream/bidirectional_stream.py
```diff
@@ -100,6 +100,9 @@ def _collect(self, response: Union[transaction_proto.Transaction.Res, transactio
         else:
             raise TypeDBClientException.of(UNKNOWN_REQUEST_ID, request_id)
 
+    def drain_errors(self) -> [RpcError]:
```
alexjpwalker, line 103: Is this a valid type hint? I've always used `List[RpcError]`

.grabl/automation.yml
flyingsilverfin: good point, forgot to bring them back!

typedb/api/connection/options.py
flyingsilverfin: We can discuss, but they make it possible to create and set options in one line, rather than spanning multiple lines which I think is still useful!

typedb/stream/bidirectional_stream.py
flyingsilverfin: HMMM i dunno, i should conform to our other styles

PR #245, opened 2022-01-05
Refactor behaviour utils

flyingsilverfin:

To match the other clients, we create a separate utils package for behaviour tests, rather than having a util file at the root of the behaviour tests.

* create a new package `//tests/behaviour/util`, and move the `wait` step into a new `util_steps.py` file
* fix a type hint for lists


typedb/stream/bidirectional_stream.py
```diff
@@ -100,7 +100,7 @@ def _collect(self, response: Union[transaction_proto.Transaction.Res, transactio
         else:
             raise TypeDBClientException.of(UNKNOWN_REQUEST_ID, request_id)
 
-    def drain_errors(self) -> [RpcError]:
+    def drain_errors(self) -> List[RpcError]:
```
flyingsilverfin, line 103: address type hinting comment

alexjpwalker (approved):

PR #247, opened 2022-01-20
Errors propagate through transaction

flyingsilverfin:

We align with Client NodeJS version 2.6.1 (mostly the work in https://github.com/vaticle/typedb-client-nodejs/pull/197), which implements a better error propagation mechanism: when an exception occurs, we store it against all the transaction's active transmit queues to retrieve whenever the user tries to perform an operation in the transaction anywhere.

* store errors received from gRPC against each receive queue
* return a new exception type, transaction is closed with errors, which throws the errors from all queues (note that this can be duplicate if there are multiple open transmit queues that have been given the same error)
* we clean up queues that are no longer needed, so we minimise the number of times the user sees the same exception


typedb/connection/transaction.py
```diff
@@ -113,7 +113,7 @@ def __exit__(self, exc_type, exc_val, exc_tb):
             return False
 
     def _raise_transaction_closed(self):
-        errors = self._bidirectional_stream.drain_errors()
+        errors = self._bidirectional_stream.get_errors()
```
flyingsilverfin, line 116: new name, aligned with client-nodejs

typedb/stream/bidirectional_stream.py
```diff
@@ -62,6 +62,9 @@ def stream(self, req: transaction_proto.Transaction.Req) -> Iterator[transaction
         self._dispatcher.dispatch(req)
         return ResponsePartIterator(request_id, self, self._dispatcher)
 
+    def done(self, request_id: UUID):
+        self._response_collector.remove(request_id)
+
```
flyingsilverfin, line 67: we now clean up completed `single`/`stream` from the response collectors, so we don't propagate and print errors for every old operation the transaction also handled but isn't active anymore!

typedb/stream/bidirectional_stream.py
```diff
@@ -78,8 +81,9 @@ def fetch(self, request_id: UUID) -> Union[transaction_proto.Transaction.Res, tr
                     raise TypeDBClientException.of(TRANSACTION_CLOSED)
                 server_msg = next(self._response_iterator)
             except RpcError as e:
-                self.close(e)
-                raise TypeDBClientException.of_rpc(e)
+                error = TypeDBClientException.of_rpc(e)
+                self.close(error)
+                raise error
```
flyingsilverfin, line 86: changed this so that we propagate our own exception instead of the gRPC error throughout the transaction queues

typedb/stream/bidirectional_stream.py
```diff
@@ -100,10 +104,10 @@ def _collect(self, response: Union[transaction_proto.Transaction.Res, transactio
         else:
             raise TypeDBClientException.of(UNKNOWN_REQUEST_ID, request_id)
 
-    def drain_errors(self) -> List[RpcError]:
-        return self._response_collector.drain_errors()
+    def get_errors(self) -> List[TypeDBClientException]:
+        return self._response_collector.get_errors()
 
-    def close(self, error: RpcError = None):
+    def close(self, error: TypeDBClientException = None):
```
flyingsilverfin, line 110: new type: use our own exception everywhere

typedb/stream/bidirectional_stream.py
```diff
@@ -127,7 +131,9 @@ def __init__(self, request_id: UUID, stream: "BidirectionalStream"):
             self._stream = stream
 
         def get(self) -> T:
-            return self._stream.fetch(self._request_id)
+            value = self._stream.fetch(self._request_id)
+            self._stream.done(self._request_id)
```
flyingsilverfin, line 135: a `Single` can immediately remove itself from the transaction stream when the user retrieves the value via `get()`

typedb/stream/response_collector.py
```diff
@@ -46,46 +46,46 @@ def new_queue(self, request_id: UUID):
...
         return self._collectors.get(request_id)
 
-    def close(self, error: Optional[RpcError]):
+    def close(self, error: Optional[TypeDBClientException]):
         with self._collectors_lock:
             for collector in self._collectors.values():
                 collector.close(error)
 
-    def drain_errors(self) -> [RpcError]:
+    def get_errors(self) -> [TypeDBClientException]:
```
flyingsilverfin, line 54: new error typing

typedb/stream/response_collector.py
```diff
@@ -46,46 +46,46 @@ def new_queue(self, request_id: UUID):
...
         errors = []
         with self._collectors_lock:
             for collector in self._collectors.values():
-                errors.extend(collector.drain_errors())
+                error = collector.get_error()
+                if error is not None:
+                    errors.append(error)
```
flyingsilverfin, line 60: get error from each collector queue

typedb/stream/response_collector.py
```diff
@@ -46,46 +46,46 @@ def new_queue(self, request_id: UUID):
...
                 return response.message
-            elif response.error:
-                raise TypeDBClientException.of_rpc(response.error)
-            else:
+            elif response.is_done() and self._error is None:
                 raise TypeDBClientException.of(TRANSACTION_CLOSED)
+            elif response.is_done() and self._error is not None:
+                raise TypeDBClientException.of(TRANSACTION_CLOSED_WITH_ERRORS, self._error)
+            else:
+                raise TypeDBClientException.of(ILLEGAL_STATE)
```
flyingsilverfin, line 78: same as in client-nodejs: we now return the recorded error, and don't put the error in the `Done` queue slot

typedb/stream/response_collector.py
```diff
@@ -94,8 +94,20 @@ class Response(Generic[R]):
...
-    def __init__(self, error: Optional[RpcError]):
-        self.error = error
+    def __init__(self):
+        pass
+
+    def is_response(self):
+        return False
+
+    def is_done(self):
+        return True
```
flyingsilverfin: is this the right way to do this in python??

typedb/stream/response_part_iterator.py
```diff
@@ -78,6 +77,7 @@ def _has_next(self) -> bool:
 
     def __next__(self) -> transaction_proto.Transaction.ResPart:
         if not self._has_next():
+            self._bidirectional_stream.done(self._request_id)
```
flyingsilverfin, line 80: we also let the transaction stream know this query stream is done so it can be removed

typedb/stream/response_collector.py
```diff
@@ -46,56 +46,71 @@ def new_queue(self, request_id: UUID):
     def get(self, request_id: UUID) -> Optional["ResponseCollector.Queue[R]"]:
         return self._collectors.get(request_id)
 
-    def close(self, error: Optional[RpcError]):
+    def remove(self, request_id: UUID):
+        with self._collectors_lock:
+            del self._collectors[request_id]
```
alexjpwalker, line 51: Something feels awkward about this - not the fault of the PR but rather the existing code.

We have a class named `ResponseCollector`, which contains an object named `_collectors`. This is simply illogical since the type of those collectors is not `ResponseCollector`.

Do you have any idea how we might improve the terminology?

typedb/stream/response_collector.py
```diff
@@ -46,56 +46,71 @@ def new_queue(self, request_id: UUID):
...
     class Queue(Generic[R]):
 
         def __init__(self):
             self._response_queue: queue.Queue[Union[Response[R], Done]] = queue.Queue()
+            self._error: TypeDBClientException = None
 
         def get(self, block: bool) -> R:
             response = self._response_queue.get(block=block)
-            if response.message:
+            if response.is_response():
```
alexjpwalker, line 75: The term `response` is getting a bit too overloaded here, can we improve on it?

typedb/stream/bidirectional_stream.py
```diff
@@ -62,6 +62,10 @@ def stream(self, req: transaction_proto.Transaction.Req) -> Iterator[transaction
         self._dispatcher.dispatch(req)
         return ResponsePartIterator(request_id, self, self._dispatcher)
 
+    def done(self, request_id: UUID):
+        print(type(self._response_collector))
```
alexjpwalker, line 66:
```suggestion
```

typedb/stream/response_collector.py
```diff
@@ -46,56 +46,71 @@ def new_queue(self, request_id: UUID):
...
+            return self._error
 
 
 class Response(Generic[R]):
 
     def __init__(self, value: R):
         self.message = value
 
+    def is_response(self):
+        return True
```
alexjpwalker, line 101: While not technically required (in either Python or TS) we did introduce a base class `QueueElement` for these two types of response (`Response` and `Done`) and given that we are now relying on the existence of a method `is_response` in each of them, we should definitely introduce a `QueueElement` base class in client-python too.

alexjpwalker (changes requested):

typedb/stream/response_collector.py
flyingsilverfin: It's a fair point, I think the idea is to see the `ResponseCollector` as a single thing - which means the naming inside is off - renamed to `_response_queues`

alexjpwalker (approved):

flyingsilverfin:
Review Summary:
We improve some terminology in the `ResponseCollector` - the collector presents itself as ONE collector. However, inside, the member variable is `collectors` -- so the question arises if the inner collectors are the actual collectors. To make this clearer, we rename the inner variable to be `response_queues`, since it represents a map to Queue objects, rather than further `Collector`s

PR #249, opened 2022-01-25
Move session pulse transmitter to the Client

alexjpwalker:

Instead of launching a background thread in each session to send pulses, we now use a single thread in the Client to send pulses for all of its sessions. This allows the client to handle a much larger number of sessions than it previously could.

**Changeset**
- Move session pulse transmitter to the Client

**Fixes**
- fixes https://github.com/vaticle/typedb-client-python/issues/222
- may reduce the chance of seeing https://github.com/vaticle/typedb-client-python/issues/243

### DIFF
--- a/typedb/connection/client.py
+++ b/typedb/connection/client.py
@@ -18,9 +18,15 @@
 #   specific language governing permissions and limitations
 #   under the License.
 #
+import sched
+import time
+from threading import Thread, Lock
 from typing import Dict
+from uuid import uuid4
 
 from grpc import Channel
+
+import typedb_protocol.common.session_pb2 as session_proto
 
 from typedb.api.connection.client import TypeDBClient
 from typedb.api.connection.options import TypeDBOptions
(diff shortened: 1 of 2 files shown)


typedb/connection/client.py
```diff
@@ -18,7 +18,11 @@
 #   specific language governing permissions and limitations
```
alexjpwalker, line 18: Instead of launching a background thread in each session to send pulses, we now use a single thread in the Client to send pulses for all of its sessions. This allows the client to handle a much larger number of sessions than it previously could.

typedb/connection/client.py
```diff
@@ -32,23 +36,30 @@
...
 
 class _TypeDBClientImpl(TypeDBClient):
+    _PULSE_INTERVAL_SECONDS = 5
 
     # TODO: Detect number of available CPUs
     def __init__(self, address: str, parallelisation: int = 2):
         self._address = address
         self._transmitter = RequestTransmitter(parallelisation)
         self._sessions: Dict[bytes, _TypeDBSessionImpl] = {}
+        self._sessions_lock = Lock()
```
alexjpwalker, line 46: We'd previously believed that `dict` was thread-safe in Python, but this has now been disproven - so we need to restrict access to our `sessions` collection using a Lock.

typedb/connection/client.py
```diff
@@ -32,23 +36,30 @@
...
         self._is_open = True
+        self._pulse_scheduler = sched.scheduler(time.time, time.sleep)
+        self._pulse = self._pulse_scheduler.enter(delay=self._PULSE_INTERVAL_SECONDS, priority=1, action=self._transmit_pulses, argument=())
+        Thread(target=self._pulse_scheduler.run, name="session_pulse_{}".format(uuid4()), daemon=True).start()
```
alexjpwalker, line 50: Schedule pulses to be transmitted every 5 seconds.

Pulses will be transmitted for all sessions that are open at the time of execution.

typedb/connection/client.py
```diff
@@ -84,5 +95,17 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+
+    def _transmit_pulses(self) -> None:
+        if not self.is_open():
+            return
+        with self._sessions_lock:
+            sessions = self._sessions.copy()
+        for session in sessions.values():
+            session.transmit_pulse()
+        self._pulse = self._pulse_scheduler.enter(delay=self._PULSE_INTERVAL_SECONDS, priority=1, action=self._transmit_pulses, argument=())
+        Thread(target=self._pulse_scheduler.run, name="session_pulse_{}".format(uuid4()), daemon=True).start()
```
alexjpwalker, line 111: Schedule a new pulse 5 seconds from now.

typedb/connection/session.py
```diff
@@ -103,41 +96,33 @@ def close(self) -> None:
...
-                try:
-                    self._scheduler.cancel(self._pulse)
-                except ValueError:  # This may occur if a pulse is in transit right now.
-                    pass
                 req = session_proto.Session.Close.Req()
                 req.session_id = self._session_id
                 try:
                     self._stub().session_close(req)
-                except RpcError:  # This generally means the session is already closed.
+                except TypeDBClientException:  # This generally means the session is already closed.
```
alexjpwalker, line 103: Fixes a bug where the client would throw if the session was already closed. It was just straight-up incorrect code - `RpcError` is never thrown by `session_close`.

typedb/connection/session.py
```diff
@@ -18,13 +18,10 @@
 # specific language governing permissions and limitations
```
alexjpwalker, line 18: We've moved the pulse scheduling logic out of this `Session` class and into `Client`.

typedb/common/concurrent/scheduled_executor.py
```diff
@@ -0,0 +1,67 @@
+#
```
alexjpwalker, line 1: Python's `scheduler` is pretty limited in what it can do. In order to make the Client code clearer, we need to introduce our own `ScheduledExecutor` class providing `schedule_at_fixed_rate`, akin to Java's `ScheduledThreadPoolExecutor`.

typedb/common/concurrent/scheduled_executor.py
```diff
@@ -0,0 +1,70 @@
...
+            self._scheduled_activity = self._scheduler.enter(delay=self._interval, priority=1, action=self._on_tick)
+
+        def _on_tick(self):
+            if self._cancelled:
+                return
+            self._schedule_run()
+            try:
+                self._action()
+            except Exception as e:
+                print(e)
```
alexjpwalker, line 65: Ensure that the next scheduled run still occurs even if an exception is thrown by the current run

PR #252, opened 2022-03-22
Query iterator only throws exeption once

flyingsilverfin:

Revert previous changes from https://github.com/vaticle/typedb-client-python/pull/247/ and https://github.com/vaticle/typedb-client-python/pull/247/, which made query queues and iterators throw the same error idempotently. However, this goes counter to standard usage of iterators and queues, which are not meant to behave idempotently (each item is only returned once, and if they have an error they should no longer be used). 

* remove idempotent error state of collectors and queues, which back query iterators
  * note that we still store the error on the transaction bidirectional stream, in case the server throws an exception when there are no query iterators active

### DIFF
--- a/typedb/stream/bidirectional_stream.py
+++ b/typedb/stream/bidirectional_stream.py
@@ -19,7 +19,7 @@
 # under the License.
 #
 from queue import Empty, Queue
-from typing import TypeVar, Iterator, Union, Generic, List
+from typing import TypeVar, Iterator, Union
 from uuid import uuid4, UUID
 
 import typedb_protocol.common.transaction_pb2 as transaction_proto
@@ -133,8 +133,7 @@ class BidirectionalStream:
             self._stream = stream
 
         def get(self) -> T:
-            value = self._stream.fetch(self._request_id)
-            return value
+            return self._stream.fetch(self._request_id)
 
 
 class RequestIterator(Iterator[Union[transaction_proto.Transaction.Req, StopIteration]]):
--- a/typedb/stream/response_collector.py
+++ b/typedb/stream/response_collector.py
@@ -24,8 +24,8 @@ from threading import Lock
 from typing import Generic, TypeVar, Dict, Optional
 from uuid import UUID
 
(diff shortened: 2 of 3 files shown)


typedb/stream/bidirectional_stream.py
```diff
@@ -133,8 +133,7 @@ def __init__(self, request_id: UUID, stream: "BidirectionalStream"):
             self._stream = stream
 
         def get(self) -> T:
-            value = self._stream.fetch(self._request_id)
-            return value
```
flyingsilverfin, line 137: inline var

typedb/stream/response_collector.py
```diff
@@ -54,26 +54,23 @@ class Queue(Generic[R]):
...
         def get(self, block: bool) -> R:
             response = self._response_queue.get(block=block)
             if response.is_value():
                 return response.value
-            elif response.is_done():
-                self._raise_transaction_closed_error()
+            elif response.is_done() and response.error is None:
+                raise TypeDBClientException.of(TRANSACTION_CLOSED)
+            elif response.is_done() and response.error is not None:
+                raise TypeDBClientException.of_rpc(response.error)
```
flyingsilverfin, line 65: go back to the old behaviour of retrieving the error from the `Done` message stored in the queue, which will be thrown exactly once. After that, this queue will always throw `transaction closed`.

typedb/stream/response_part_iterator.py
```diff
@@ -74,9 +74,7 @@ def _has_next(self) -> bool:
             raise TypeDBClientException.of(ILLEGAL_STATE)
 
     def __next__(self) -> transaction_proto.Transaction.ResPart:
-        if self._bidirectional_stream.get_error() is not None:
```
flyingsilverfin, line 77: the error will come from the backing `Queue`, rather than the bidirectional stream.

alexjpwalker:
> Revert previous changes from https://github.com/vaticle/typedb-client-python/pull/247 and https://github.com/vaticle/typedb-client-python/pull/247
@flyingsilverfin lol

alexjpwalker (approved):
Seems identical to the Java client PR 👍

Issue #265, opened 2022-10-13
Easier Value Retrieval


suciokhan:
Please replace every line in curly brackets { like this } with an appropriate description, and remove this line.

## Problem to Solve

Currently retrieving the values from attributes queried for is quite cumbersome. For example, if I want to get a series of entities and the values of their attributes that meet specific criteria, I have to run a separate .get(var) for each variable I write in the query. If there were a simple way to return the results in a dictionary/list-of-dictionaries type of format containing all the key:value items queried for it would greatly enhance productivity. As it stands, I spend most of my time figuring out how to transform the data into a usable format rather than exploiting the data from these queries.

## Current Workaround

Make a separate list object for every variable I want information on; which requires different lists for each query depending on how may variables I'm querying for. I then make an empty list, and append a dictionary that contains the values for a single result by iterating over a range the length of one of the lists, using the indices for value retrieval and have to manually create the key names.

For example:
```
items = []
for i in range(0,len(wid)):
    items.append({
        'researcher_id':aid[i],
        'researcher_name':aname[i],
        'work_id':wid[i],
        'work_title':wtitle[i],
        'cited_by_count':cbc[i],
        'institution_name':iname[i],
        'institution_city':icity[i]
    })
```

## Proposed Solution

Make a function that easily lets you squirt the results into this kind of a format without requiring the user to manually create lists/dictionaries, etc. for holding the values and then having to transform them.

Something like:
```
with session.transaction(Transactiontype.READ) as read_transaction:
    results_json = read_transaction.query().match(query)

results_json = [{researcher_id: 'abcd', 'researcher_name': "Bob", 'work_id':'12324', ... }, {...}]
```

## Additional Information

NA

alexjpwalker:
@suciokhan : It sounds like you're looking for [`concept_map.map()`](https://docs.vaticle.com/docs/client-api/python#retrieve-a-mapping-of-variables-to-concepts)

```py
results_maps = [answer.map() for answer in read_transaction.query().match(query)]

**Comment by suciokhan** [comment]
Hmmm...that's close, but it doesn't have the values...

```
with TypeDB.core_client("127.0.0.1:1729") as client:
    with client.session("database", SessionType.DATA) as session:
        with session.transaction(TransactionType.READ, options) as read_transaction:
            # Do a query
            #answer_iterator = read_transaction.query().match(query)
            results_maps = [answer.map() for answer in read_transaction.query().match(query)]
```
gives me:
```
[...]

PR #266, opened 2022-11-01
Remove calls to `grpc.channel.close()`, which could cause segfaults

### LINKED ISSUE #239: GRPC memory leak from opening session using empty with block
## Description

A basic session open/close example in docs has broken as of upgrading from client 2.2.0 to 2.4.0. The breakage comes from GRPC, seemingly because a resource wasn't closed properly. It's very suspicious that this only breaks for an empty `with` block, so I have named the issue as such.

## Environment

1. OS (where TypeDB server runs): Factory (not reproducible locally on Mac OS 10)
2. TypeDB version (and platform): TypeDB 2.5
3. TypeDB client-python version: client-python 2.5
4. Python version: 3.6

## Reproducible Steps

Steps to create the smallest reproducible scenario:

Run the docs tests in Factory, and most of the time the Python tests fail in this way:
```
INFO: Found 1 target...
1255
[0 / 1] [Prepa] BazelWorkspaceStatusAction stable-status.txt
1256
Target //test/example/python:social-network up-to-date:
1257
  bazel-bin/test/example/python/social_network_quickstart_query.py
1258
  bazel-bin/test/example/python/social_network_python_client_a.py
1259
  bazel-bin/test/example/python/social_network_python_client_b.py
1260
  bazel-bin/test/example/python/social_network_python_client_c.py
1261
  bazel-bin/test/example/python/social_network_python_client_d.py
1262
  bazel-bin/test/example/python/social-network
1263
INFO: Elapsed time: 0.842s, Critical Path: 0.51s
1264
INFO: 5 processes: 1 remote cache hit, 4 internal.
1265
INFO: Build completed successfully, 5 total actions
1266
INFO: Running command line: external/bazel_tools/tools/test/test-setup.sh test/example/python/social-network
1267
INFO: Build completed successfully, 5 total actions
1268
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
1269
Executing tests from //test/example/python:social-network
1270
-----------------------------------------------------------------------------
1271
test_social_network_python_client_a (__main__.SocialNetworkTest) ... ok
1272
test_social_network_python_client_b (__main__.SocialNetworkTest) ... E1026 09:05:48.592246686    4855 metadata.cc:253]            WARNING: 1 metadata elements were leaked
1273
E1026 09:05:48.592321591    4855 metadata.cc:260]            mdelem 'user-agent' = 'grpc-python/1.38.0 grpc-c/16.0.0 (linux; chttp2)'
1274
E1026 09:05:48.592333692    4855 metadata.cc:253]            WARNING: 1 metadata elements were leaked
1275
E1026 09:05:48.592339292    4855 metadata.cc:260]            mdelem ':authority' = 'localhost:1729'
```

Investigating, `test_social_network_python_client_b` seems to be giving a warning of a leak.I presume that although this says `"WARNING"`, given the lack of subsequent output, this is our culprit. The user-agent is `grpc-python` and the error codes seem consistent with issues on grpc: https://github.com/grpc/grpc/issues/7121

The snippet of the docs being tested, `social_network_python_client_b`, is trivial and is as follows:

```python
from typedb.client import *

with TypeDB.core_client("localhost:1729") as client:
    with client.session("social_network", SessionType.DATA) as session:
        ## session is open
        pass
    ## session is closed
## client is closed
```

It looks like this has been broken since commit `fb4fbfed0c97f35ccd7984e637f6f6b4c2fcf023`, which upgraded the python client dependency from `2.2.0` to `2.4.0`.

Note that this issue is non-deterministic, I just re-ran the tests on Factory and they passed. I was also not able to reproduce this locally.

## Expected Output

Session should close with no effect!

## Actual Output

Memory leak and error from GRPC.

alexjpwalker:

We no longer call `close` on any of our gRPC Channels. This fixes possible segfaults caused by resources being deallocated while they are still in use.

Users in a wide variety of scenarios had reported intermittent crashes, often accompanied by warnings saying "1 metadata element(s) leaked". The logs would look similar to the following:
```
7786| E1031 17:59:35.477338904 207 metadata.cc:253] WARNING: 1 metadata elements were leaked
7787| E1031 17:59:35.477409424 207 metadata.cc:260] mdelem ':authority' = 'typedb-cluster-1.typedb-cluster:1729'
7788| [mutex.cc : 435] RAW: Lock blocking 0x563cbf829fc0 @
7789| [mutex.cc : 1908] RAW: Check (v & (kMuWait | kMuWrWait)) != kMuWrWait failed: Lock: Mutex corrupt: waiting writer with no waiters: 0x563cbf815670
```
The same issue has also been reported in https://github.com/googleads/google-ads-python/issues/384, and a fix was suggested in:
- https://github.com/grpc/grpc/issues/19235

From this issue we infer that `Channel.close` is not behaving nicely in gRPC Python, and can cause resources to be deallocated while they are still in use. As gRPC itself uses native C libraries, this results in segfaults and crashes. We determine that the best course of action is to not close the Channel ourselves.

We've simply deleted the 3 places in our code that called `close` on a gRPC `Channel`. It has passed our CI tests and fixed user-reported issues, and it is what the gRPC maintainers themselves appear to recommend in the linked `grpc` issue.

- closes #239 
- see #265


.factory/automation.yml
```diff
@@ -32,201 +32,164 @@ build:
...
+      image: vaticle-ubuntu-22.04
       command: |
         bazel run @vaticle_dependencies//factory/analysis:dependency-analysis
   correctness:
     build:
-      image: vaticle-ubuntu-21.04
+      image: vaticle-ubuntu-22.04
       type: foreground
       command: |
-        pyenv install 3.7.12
```
alexjpwalker, line 43: We had to make a number of changes to the way we load Python and Pip and install packages in order to work in the new Factory CI machines.

typedb/connection/cluster/server_client.py
```diff
@@ -63,7 +63,6 @@ def _new_channel(self) -> grpc.Channel:
 
     def close(self) -> None:
         super().close()
-        self._channel.close()
```
alexjpwalker, line 66: See PR description for the full explanation of why we do this. In brief, explicitly calling `close` on a gRPC Channel in Python can result in resources being deallocated while they are still being used, which in turn may cause gRPC's native C libraries to segfault. Not closing the channel is the approach recommended in an issue in the gRPC repository itself.

PR #267, opened 2022-11-14
Upgrade to 2.12.0

flyingsilverfin:

We synchronise client-python with client-java 2.12.0, which includes changes to the protocol up to 2.12.0.

* Upgrade dependencies to protocol and client 2.12.0, and typedb-core and cluster to 2.13.0
* Integrate protocol changes, notably getting rid of `is_abstract()` and using the boolean packaged into the type messages, as well as introducing new capabilities for `RoleType`
* Update behaviour steps to implement new tests


typedb/common/rpc/request_builder.py
```diff
@@ -426,9 +420,39 @@ def role_type_get_relation_types_req(label: Label):
...
+def role_type_get_player_types_req(label: Label):
+    req = concept_proto.Type.Req()
+    req.role_type_get_player_types_req.CopyFrom(concept_proto.RoleType.GetPlayerTypes.Req())
+    return type_req(req, label)
+
+
+def role_type_get_player_types_explicit_req(label: Label):
     req = concept_proto.Type.Req()
-    req.role_type_get_players_req.CopyFrom(concept_proto.RoleType.GetPlayers.Req())
+    req.role_type_get_player_types_explicit_req.CopyFrom(concept_proto.RoleType.GetPlayersTypesExplicit.Req())
```
alexjpwalker, line 431: Seems to be correctly named `GetPlayerTypesExplicit` in `typedb-protocol`. Let's do a quick sweep for typos, they don't cause the build to fail in Python.
```suggestion
    req.role_type_get_player_types_explicit_req.CopyFrom(concept_proto.RoleType.GetPlayerTypesExplicit.Req())
```

typedb/concept/proto/concept_proto_reader.py
```diff
@@ -84,7 +84,7 @@ def thing_type(proto_type: concept_proto.Type):
     elif proto_type.encoding == concept_proto.Type.Encoding.Value("ATTRIBUTE_TYPE"):
         return attribute_type(proto_type)
     elif proto_type.encoding == concept_proto.Type.Encoding.Value("THING_TYPE"):
-        return _ThingType(Label.of(proto_type.label), proto_type.root)
+        return _ThingType(Label.of(proto_type.label), proto_type.isRoot, proto_type.isAbstract)
```
alexjpwalker, line 87: Are these really camel cased? This looks like a typo to me.
```suggestion
        return _ThingType(Label.of(proto_type.label), proto_type.is_root, proto_type.is_abstract)
```
Of course, we know the properties of the root thing type, so this is also an option ...
```suggestion
        return _ThingType(label = Label.of(proto_type.label), is_root = True, is_abstract = True)
```

alexjpwalker: Looks like this same typo has been made in a bunch of other places in the PR.

typedb/concept/type/type.py
```diff
@@ -37,11 +37,12 @@
 
 class _Type(Type, _Concept, ABC):
 
-    def __init__(self, label: Label, is_root: bool):
+    def __init__(self, label: Label, is_root: bool, is_abstract):
```
alexjpwalker, line 40:
```suggestion
    def __init__(self, label: Label, is_root: bool, is_abstract: bool):
```

typedb/common/rpc/request_builder.py
```diff
@@ -555,6 +579,11 @@ def thing_type_get_instances_req(label: Label):
     req.thing_type_get_instances_req.CopyFrom(concept_proto.ThingType.GetInstances.Req())
     return type_req(req, label)
 
+def thing_type_syntax_req(label: Label):
+    req = concept_proto.Type.Req()
+    req.thing_type_syntax_req.CopyFrom(concept_proto.ThingType.GetSyntax.Req())
```
alexjpwalker, line 584: Typo, see protocol + rename method accordingly
```suggestion
    req.thing_type_get_syntax_req.CopyFrom(concept_proto.ThingType.GetSyntax.Req())
```

alexjpwalker (changes requested):

typedb/concept/proto/concept_proto_reader.py
flyingsilverfin: is actually correct :D

alexjpwalker (approved):

typedb/concept/proto/concept_proto_reader.py
alexjpwalker: @flyingsilverfin sorry, I should chased this down harder. Root cause was a bug in `typedb_protocol`. fixed by:
- https://github.com/vaticle/typedb-protocol/pull/175

PR #282, opened 2023-03-02
Fix duplicate requirements specification

### LINKED ISSUE #219: Project requirements are specified in (at least) two places
## Problem to Solve

`requirements_dev.txt` was added in order to provide a convenient way of installing dev dependencies that we only need when developing Client Python, not when you simply want to use it. This is akin to `devDependencies` in `package.json` of client-nodejs. It includes a test framework, `behave`, and an assertion framework `PyHamcrest`.

But it's an ugly solution. We now have to specify the version of, for example, `grakn-protocol` in 2 places - `requirements.txt` and `requirements_dev.txt`.

## Proposed Solution

I originally created this split so that we would only have to run one command: `pip install -r requirements_dev.txt`. But perhaps running two commands is a better option (i.e. yank the prod requirements out of requirements_dev, hence requiring that we run both.)

After all, this is also something Bazel build _should_ be able to do automatically for us. I've noticed that we do have a `pip_install` rule in our `WORKSPACE`, and yet it doesn't seem to actually install the packages from what I can tell. I always have to run `pip install` myself manually after the Bazel build.

adammichaelwood:

This PR solves issue #219 by including `requirements.txt` inside of `requirements_dev.txt`, so that the requirements are no longer duplicated. Going forward, developers will no longer need to remember to copy changes from `requirements.txt` to `requirements_dev.txt`.

This PR adds a reference (`-r requirements.txt`) to `requirements_dev.txt`, and cleans up the instructions around it. See the [pip install Requirements File Format](https://pip.pypa.io/en/stable/reference/requirements-file-format/) for details on this feature.

closes #219

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,7 +19,11 @@
 # under the License.
 #
 
-## To install all dependencies, run: pip install -r requirements.txt
+## To install primary (not dev) dependencies, run: 
+#  pip install -r requirements.txt
+
+## To install ALL (including dev) dependencies, run:
+#  pip instal -r requirements_dev.txt
 
 
 ## Configuration options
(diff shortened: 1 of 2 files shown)


flyingsilverfin:
Thank you for the contribution!

PR #283, opened 2023-03-03
Cluster password policy and user management support

dmitrii-ubskii:

Add functionality relevant to the changes made in https://github.com/vaticle/typedb-cluster/pull/456:
* Optional password expiry in days for each user.
* The ability for admins to set passwords of other users
* User password update, which requires old and new password

We've added the functionality that is being added in a recent TypeDB Cluster pull request:
* Support updating password now requiring both the old and the new password.
* Add support for retrieving the user password expiry in days. This is optional, depending on whether password expiration is enabled on the server.
* Allow updating password as the admin requiring only a new password.

### DIFF
--- a/.factory/automation.yml
+++ b/.factory/automation.yml
@@ -40,6 +40,11 @@ build:
       image: vaticle-ubuntu-22.04
       type: foreground
       command: |
+        export PATH="$HOME/.local/bin:$PATH"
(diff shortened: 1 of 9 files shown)


flyingsilverfin (approved):

flyingsilverfin:
Good stuff!

PR #288, opened 2023-04-05
Define user password expiry in terms of seconds

jamesreprise:

We've redefined the way we expose how long until a password expires in seconds, rather than days as was done previously.

* We've changed `password_expiry_days` to `password_expiry_seconds`.
* We implement the newly updated BDD steps to do with opening and closing connections, and prepare other steps for booting and configuring TypeDB for use once the TypeDB runner is available in Python


jamesreprise:
Still waiting on 2.17.0 release from both servers. Also need to update behaviour, though this is likely to be time intensive.

flyingsilverfin (approved):

PR #293, opened 2023-05-16
Support generalised annotations and uniqueness

flyingsilverfin:

We update the Typedb Protocol and TypeQL to the latest versions, which support the uniqueness annotation and generalised `Type` APIs. These generalised APIs allow querying by a set of annotations, rather than just boolean = `true|false`.

For example:
`AttributeType.get_owners(onlyKey: boolean)` becomes `AttributeType.get_owners(annotations: Set["Annotation"])`

All usages of boolean flags to indicate key-ness should be replaced by using `ThingType.AnnotationEnum.KEY`. The new `@unique` annotation is available as `ThingType.AnnotationEnum.KEY`.

* Update dependencies to include generalised annotations in the language and protocol. We also pull the `behaviour` repository to include tests for generalised annotations
* We reflect the changes to the behaviour test steps in the implementation of BDD steps, in particular with extensions for generalised annotations
* Update Concept API methods to remove the `*_key` boolean and replace it with a set of flags from `ThingType.Annotation`

Related protocol changes: https://github.com/vaticle/typedb-protocol/pull/182 and https://github.com/vaticle/typedb-protocol/commit/de4f5b4d31c3ff06d62da47d2a57605c4e3d4b29

Related core changes: https://github.com/vaticle/typedb/pull/6775 , https://github.com/vaticle/typedb/commit/e24cfa981570e8ba5db392f274c9c8cedc6764cf , and https://github.com/vaticle/typedb/commit/b16cdb0be71968588f58bccd1088529cdd003063

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -35,7 +35,7 @@
 
 ## Dependencies
 
-typedb-protocol==de4f5b4d31c3ff06d62da47d2a57605c4e3d4b29
+typedb-protocol==2.18.0.dev0
 grpcio>=1.43.0,<2
 protobuf>=3.15.6,<4
 parse==1.18.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -35,6 +35,6 @@
 
 ## Dependencies
 
-typedb-protocol==2.17.0
+typedb-protocol==de4f5b4d31c3ff06d62da47d2a57605c4e3d4b29
 grpcio>=1.43.0,<2
-protobuf>=3.15.5,<4
+protobuf>=3.15.6,<4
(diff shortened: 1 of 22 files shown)


requirements.txt
```diff
@@ -35,6 +35,7 @@
 
 ## Dependencies
 
-typedb-protocol==2.17.0
+typedb-protocol==2.18.0.dev0
```
flyingsilverfin, line 38: Python's new conventions means we have to use properly named pre-releases. Until python can be a rust-wrapped client, we will use this convention and release snapshots for Python manually.

PR #296, opened 2023-05-25
Introduce Values to support expressions

flyingsilverfin:

Introduce the 'Value' type, which is returned as the result of an expression's computation. This change follows from https://github.com/vaticle/typeql/pull/260, which outlines the capabilities of the new expression syntax.

Values (representing any of Long, Double, Boolean, String, or DateTime) are returned as part of `ConceptMap` answers and are subtypes of `Concept` for the time being. Their main API is made of the `.get_value()` method and `.get_value_vype()` method, along with all the standard safe downcasting methods to convert a `Concept` into a `Value`, using `Concept.is_value()` and `Concept.as_value()`.

We also move the import location of `ValueType` from `attribute_type.py` to `concept.py`.

* Introduces the `Value` concept and the required `ValueImpl` that implements it
* Refactor `ValueType` to no longer live within `AttributeType` - now it exists in `Concept.ValueType`
* Updates the test framework for tests involving values, including the new `ExpressionTest` behaviour scenarios, which we also add to CI

### DIFF
--- a/typedb/api/concept/value/value.py
+++ b/typedb/api/concept/value/value.py
@@ -1,0 +1,124 @@
+#
+# Copyright (C) 2022 Vaticle
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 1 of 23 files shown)


typedb/api/concept/value/value.py
```diff
@@ -0,0 +1,124 @@
...
+from abc import ABC, abstractmethod
+from datetime import datetime
+from typing import TYPE_CHECKING, Mapping, Union
+
+from typedb.api.concept.concept import Concept, ValueType
+from typedb.api.connection.transaction import TypeDBTransaction
+from typedb.common.exception import TypeDBClientException, VALUE_HAS_NO_REMOTE
+
+if TYPE_CHECKING:
+    pass
```
alexjpwalker, line 30: We should just delete this, right? It's a no-op
```suggestion
```

alexjpwalker (approved):

typedb/api/concept/value/value.py
flyingsilverfin: Sure

PR #297, opened 2023-05-30
Protocol versioning

flyingsilverfin:

We use a new protocol API to perform a "connection open". This API does server-side protocol version compatibility checks, and replaces our previous need to get all databases to check that the connection is available.

This API is called transparently during the construction of both the Core and Cluster clients.

* Refactor Stubs to call `connectionOpen` in their constructors
* Throw a useful message if the server returns an `unimplemented` error

### DIFF
--- a/typedb/common/exception.py
+++ b/typedb/common/exception.py
@@ -38,7 +38,9 @@ class TypeDBClientException(Exception):
 
     @staticmethod
     def of_rpc(rpc_error: Union[RpcError, Call]) -> "TypeDBClientException":
-        if rpc_error.code() in [StatusCode.UNAVAILABLE, StatusCode.UNKNOWN] or "Received RST_STREAM" in str(rpc_error):
+        if rpc_error.code() in [StatusCode.UNIMPLEMENTED]:
+            return TypeDBClientException(msg=RPC_METHOD_UNAVAILABLE, cause=rpc_error.details())
+        elif rpc_error.code() in [StatusCode.UNAVAILABLE, StatusCode.UNKNOWN] or "Received RST_STREAM" in str(rpc_error):
             return TypeDBClientException(msg=UNABLE_TO_CONNECT, cause=rpc_error)
         elif rpc_error.code() is StatusCode.INTERNAL and "[RPL01]" in str(rpc_error):
             return TypeDBClientException(msg=CLUSTER_REPLICA_NOT_PRIMARY, cause=None)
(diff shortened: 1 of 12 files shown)


typedb/common/exception.py
```diff
@@ -38,7 +38,9 @@ def __init__(self, msg: Union["ErrorMessage", str], cause: BaseException = None,
 
     @staticmethod
     def of_rpc(rpc_error: Union[RpcError, Call]) -> "TypeDBClientException":
-        if rpc_error.code() in [StatusCode.UNAVAILABLE, StatusCode.UNKNOWN] or "Received RST_STREAM" in str(rpc_error):
+        if rpc_error.code() is StatusCode.UNIMPLEMENTED:
```
flyingsilverfin, line 41: New user-friendly error parsing for when the server is missing a method

typedb/connection/client.py
```diff
@@ -42,11 +43,13 @@ def __init__(self, address: str, parallelisation: int = 2):
         self._transmitter = RequestTransmitter(parallelisation)
         self._sessions: Dict[bytes, _TypeDBSessionImpl] = {}
         self._sessions_lock = Lock()
-        self._is_open = True
         self._pulse_executor = ScheduledExecutor()
         self._pulse_executor.schedule_at_fixed_rate(interval=self._PULSE_INTERVAL_SECONDS, action=self._transmit_pulses)
 
     def session(self, database: str, session_type: SessionType, options=None) -> _TypeDBSessionImpl:
+        if not self.is_open():
```
flyingsilverfin, line 50: since opening a client is now more intricate, we validate the client is fully opened before allowing sessions to be opened

typedb/connection/cluster/server_client.py
```diff
@@ -40,6 +40,7 @@ def __init__(self, address: str, credential: TypeDBCredential, parallelisation:
             self._channel_credentials = grpc.ssl_channel_credentials()
         self._channel, self._stub = self.new_channel_and_stub()
         self._databases = _TypeDBDatabaseManagerImpl(self.stub())
+        self._is_open = True
```
flyingsilverfin, line 43: we move the `is_open = True` down from the parent constructor into each sub-constructor

typedb/connection/cluster/stub.py
```diff
@@ -47,8 +47,10 @@ def __init__(self, channel: Channel, credential: TypeDBCredential):
         self._channel = channel
         self._stub = core_service_proto.TypeDBStub(channel)
         self._cluster_stub = cluster_service_proto.TypeDBClusterStub(channel)
+
         self._token = None
         try:
+            self._stub.connection_open(connection_open_req())
```
flyingsilverfin, line 53: the `stub` directly calls `open` on the server when being constructed.

flyingsilverfin:
@alexjpwalker  maybe you can do a post-review, I will merge it for now to proceed with release

PR #304, opened 2023-08-02
Implement native calls from Python Driver to Rust Driver

dmikhalin:

We replace the direct implementation of the client side of the TypeDB protocol in Python with native calls to the Rust client over FFI.

- merge `TypeDBClient` and `ClusterClient` into single `Client`, following the protocol merge;
- remove `Remote` variants of `Concept`s; Concept API calls now require a `Transaction` to be provided;
- make `Value` "generic" over its value type, which allows us to collapse all the `Attribute.<Type>`s into a single `Attribute` class;
- associate each top-level class (`Client`, `Session`, `Concept`, etc.) with its native representation on the Rust side;
- replace Python method implementations with calls using the native representation over FFI.


typedb/concept/thing/attribute.py
```diff
@@ -18,223 +18,33 @@
...
-    def get_value(self):
-        return self._value
-
-    def as_remote(self, transaction):
-        return _RemoteDateTimeAttribute(transaction, self.get_iid(), self.is_inferred(), self.get_type(), self.get_value())
+    def get_owners(self, transaction: _Transaction, owner_type: Optional[_ThingType] = None) -> Iterator[Any]:
+        return (_Thing.of(item) for item in Streamer(attribute_get_owners(self.native_transaction(transaction),
+                                                                          self.native_object,
+                                                                          owner_type.native_object if owner_type else None
+                                                                          ), concept_iterator_next))
```
dmitrii-ubskii, line 50: Some significant rightward drift here.

typedb/concept/thing/entity.py
```diff
@@ -19,45 +19,15 @@
...
-        return _Entity(concept_proto_reader.iid(thing_proto.iid), thing_proto.inferred, concept_proto_reader.type_(thing_proto.type))
-
-    def get_type(self) -> "EntityType":
-        return self._type
-
-    def as_remote(self, transaction):
-        return _RemoteEntity(transaction, self._iid, self.is_inferred(), self.get_type())
-
-    def as_entity(self) -> "Entity":
-        return self
```
dmitrii-ubskii, line 47: `as_foo()` is not overridden on many of the concepts as far as I can see. Won't they just raise an exception based on the `Concept` implementation?

typedb/concept/type/entity_type.py
```diff
@@ -19,35 +19,79 @@
...
+class _EntityType(EntityType, thing_type._ThingType):
 
-    def as_entity_type(self) -> "EntityType":
-        return self
+    # def __init__(self, concept: Concept):
+    #     super(_ThingType).__init__(concept)
 
+    # @staticmethod
+    # def of(type_proto: concept_proto.Type):
+    #     return _EntityType(Label.of(type_proto.label), type_proto.is_root, type_proto.is_abstract)
```
dmitrii-ubskii, line 48: Everything commented out needs to go

typedb/concept/type/thing_type.py
```diff
@@ -18,103 +18,192 @@
...
+    from typedb.concept.type.role_type import _RoleType
+    from typedb.connection.transaction import _Transaction
+
+
+class _ThingType(ThingType, _Type, ABC):
+
+    @staticmethod
+    def of(concept: NativeConcept):
+        from typedb.concept.type import attribute_type, entity_type, relation_type
+
```
dmitrii-ubskii, line 55: Why not import at the top, and why not import the classes? We already need `_AttributeType` for type hints anyway

typedb/concept/type/thing_type.py
```diff
@@ -18,103 +18,192 @@
...
+    def get_plays(self, transaction: _Transaction, transitivity: Transitivity = Transitivity.Transitive) -> Iterator[_RoleType]:
+        from typedb.concept.type import role_type
+
+        return (role_type._RoleType(item) for item in
+                Streamer(thing_type_get_plays(
+                    self.native_transaction(transaction),
+                    self.native_object,
+                    transitivity.value
+                ), concept_iterator_next)
```
dmitrii-ubskii, line 148: Is there a difference between this and `map(role_type._RoleType, Streamer(...))`? The `map()` version seems a litlle tighter than the generator.

typedb/connection/session.py
```diff
@@ -18,110 +18,70 @@
...
-            options = TypeDBOptions.core()
-        self._client = client
-        self._address = client.address()
-        self._session_type = session_type
+            options = Options()
+        self._type = session_type
         self._options = options
-        self._rw_lock = ReadWriteLock()
-        self._database = _TypeDBDatabaseImpl(stub=self._stub(), name=database)
+        db = database.native_object
```
dmitrii-ubskii, line 46:
```suggestion
        native_database = database.native_object
```

typedb/connection/user.py
dmitrii-ubskii, line 1: `user` should move to a top-level user package now, here and in `api`

tests/behaviour/background/environment_base.py
```diff
@@ -43,10 +44,8 @@ def before_scenario(context: Context):
     context.put = lambda var, thing: _put_impl(context, var, thing)
     context.get_thing_type = lambda root_label, type_label: _get_thing_type_impl(context, root_label, type_label)
     context.clear_answers = lambda: _clear_answers_impl(context)
-    context.option_setters = {
-        "session-idle-timeout-millis": lambda option, value: option.set_session_idle_timeout_millis(int(value)),
-        "transaction-timeout-millis": lambda option, value: option.set_transaction_timeout_millis(int(value)),
-    }
+    context.option_setters = {option: lambda options, value: setattr(options, option.replace("-", "_"), value) for option in
+                              ["session-idle-timeout-millis", "transaction-timeout-millis"]}
```
dmitrii-ubskii, line 48: We lost `int(value)` in this change

tests/behaviour/concept/serialization/json/BUILD
```diff
@@ -48,6 +48,8 @@ typedb_behaviour_py_test(
         "//tests/behaviour/util:util",
         "//tests/behaviour/config:parameters",
         "//tests/behaviour/background",
+        "@vaticle_typedb_driver_java//rust:typedb_client_python",
+        "//:python-lib",
```
dmitrii-ubskii, line 52: Why are we depending on both? Shouldn't this just be `//:python-lib`?

tests/behaviour/concept/thing/relation/relation_steps.py
```diff
@@ -87,51 +86,52 @@ def step_impl(context: Context, var1: str, role_label: str, var2: str):
...
     try:
-        relation.as_remote(context.tx()).add_player(
-            relation.get_type().as_remote(context.tx()).get_relates(role_label),
+        relation.add_player(
+            context.tx(),
+            relation.get_type().get_relates(context.tx(), role_label),
             context.get(var2))
         assert False;
-    except TypeDBClientException:
+    except (TypeDBClientException, RuntimeError):
```
dmitrii-ubskii, line 94: Can this be more specific than `RuntimeError`?

tests/behaviour/concept/type/thingtype/BUILD
```diff
@@ -29,35 +29,37 @@ py_library(
...
+#        "//tests/behaviour/util:util",
+#        "//tests/behaviour/config:parameters",
+#        "//tests/behaviour/background",
+#        "@vaticle_typedb_driver_java//rust:typedb_client_python",
+#        "//:python-lib",
+#    ],
+#    native_typedb_artifact = "//tests:native-typedb-artifact",
+#    native_typedb_cluster_artifact = "//tests:native-typedb-cluster-artifact",
+#    size = "medium",
+#)
```
dmitrii-ubskii, line 62: !

tests/behaviour/context.py
```diff
@@ -46,22 +46,22 @@ class Context(behave.runner.Context):
...
-        self.sessions_parallel_to_transactions_parallel: Dict[Future[TypeDBSession], List[TypeDBTransaction]] = {}
-        self.session_options: Optional[TypeDBOptions] = None
-        self.transaction_options: Optional[TypeDBOptions] = None
-        self.things: Dict[str, Thing] = {}
-        self.answers: Optional[List[ConceptMap]] = None
+        self.client: Optional[Client] = None
+        self.sessions: list[Session] = []
+        self.sessions_to_transactions: dict[Session, list[Transaction]] = {}
+        # self.sessions_parallel: list[Future[TypeDBSession]] = []
+        # self.sessions_parallel_to_transactions_parallel: dict[Future[TypeDBSession], list[Transaction]] = {}
```
dmitrii-ubskii, line 53:
```suggestion
```

dmitrii-ubskii (changes requested):
Looks good! These changes definitely make the client a lot tidier.

Left a few style comments.

typedb/api/concept/thing/relation.py
```diff
@@ -18,49 +18,47 @@
...
         pass
 
     @abstractmethod
-    def get_players(self, role_types: Optional[List["RoleType"]] = None) -> Iterator[Thing]:
+    def get_players_by_role_type(self, transaction: Transaction, *role_types: RoleType) -> Iterator[Thing]:
         pass
 
     @abstractmethod
-    def get_players_by_role_type(self) -> Dict["RoleType", List[Thing]]:
+    def get_players(self, transaction: Transaction) -> dict[RoleType, list[Thing]]:
```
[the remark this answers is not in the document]
dmikhalin: Importing those from typing is deprecated since 3.9 ([https://peps.python.org/pep-0585/](https://peps.python.org/pep-0585/))

typedb/api/concept/concept.py
```diff
@@ -18,26 +18,14 @@
...
-    from typedb.api.concept.thing.relation import Relation, RemoteRelation
-    from typedb.api.concept.thing.thing import Thing, RemoteThing
-    from typedb.api.concept.type.attribute_type import AttributeType, RemoteAttributeType
-    from typedb.api.concept.type.entity_type import EntityType, RemoteEntityType
-    from typedb.api.concept.type.relation_type import RelationType, RemoteRelationType
-    from typedb.api.concept.type.role_type import RoleType, RemoteRoleType
-    from typedb.api.concept.type.thing_type import ThingType, RemoteThingType
-    from typedb.api.concept.type.type import Type, RemoteType
-    from typedb.api.connection.transaction import TypeDBTransaction
+from typedb.api.concept import thing, type, value
```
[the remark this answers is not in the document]
dmikhalin: We need it to avoid circular imports.

typedb/concept/thing/entity.py
dmikhalin: Yeah, we should put them back.

typedb/api/answer/numeric_group.py
```diff
@@ -18,10 +18,14 @@
 # specific language governing permissions and limitations
 # under the License.
 #
+
+from __future__ import annotations
```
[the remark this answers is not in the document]
dmikhalin: Yes, we import not all classes used in type annotations.

tests/behaviour/concept/thing/relation/relation_steps.py
dmikhalin: There is no more specific exceptions in SWIG's list of possible python exceptions.

typedb/api/connection/client.py
```diff
@@ -18,55 +18,50 @@
...
-from typedb.api.connection.user import UserManager, User
+if TYPE_CHECKING:
+    from typedb.api.connection.database import DatabaseManager
+    from typedb.api.connection.options import Options
+    from typedb.api.connection.session import Session, SessionType
+    from typedb.api.connection.user import UserManager, User
 
 
-class TypeDBClient(ABC):
+class Client(ABC):
```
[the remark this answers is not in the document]
dmikhalin: Renamed back.

tests/behaviour/concept/serialization/json/BUILD
dmikhalin: Yes, it should.

typedb/api/concept/concept.py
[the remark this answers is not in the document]
dmikhalin: Method `.of` became `concept_of` function in the separate `concept_factory.py` file.

typedb/concept/thing/relation.py
```diff
@@ -18,78 +18,55 @@
...
+from typedb.api.concept.thing.relation import Relation
 from typedb.api.concept.type.role_type import RoleType
-from typedb.common.rpc.request_builder import relation_add_player_req, relation_remove_player_req, \
-    relation_get_players_req, relation_get_players_by_role_type_req, relation_get_relating_req
-from typedb.concept.proto import concept_proto_builder, concept_proto_reader
-from typedb.concept.thing.thing import _Thing, _RemoteThing
+from typedb.common.streamer import Streamer
+from typedb.concept.concept_factory import relation_type_of, role_type_of, thing_of
+from typedb.concept.thing.thing import _Thing
 from typedb.concept.type.role_type import _RoleType
```
dmikhalin, line 30: We don't need this import here, because we use `role_type_of()`. But without direct importing of `role_type.py` Bazel think that we don't need it and removes this file. One should investigate this behaviour.

typedb/concept/concept_factory.py
dmikhalin, line 1: We added this file to have all inconvenient module imports in one place.

typedb/concept/type/thing_type.py
```diff
@@ -18,103 +18,166 @@
...
+    def is_abstract(self) -> bool:
+        return thing_type_is_abstract(self.native_object)
 
-class _RemoteThingType(_RemoteType, RemoteThingType):
+    def get_label(self) -> Label:
+        return Label.of(thing_type_get_label(self.native_object))
 
-    def as_remote(self, transaction):
-        return _RemoteThingType(transaction, self.get_label(), self.is_root(), self.is_abstract())
+    def delete(self, transaction: _Transaction) -> None:
```
flyingsilverfin, line 63: What happens if `delete()` is called, and then the user calls a method on the Python instance which already has been deleted in the Rust layer, so the native object is 'invalid' (if that's how we've designed it)

typedb/connection/database.py
```diff
@@ -19,25 +19,95 @@
...
+    def __init__(self, database: NativeDatabase):
+        self._native_object = database
+        self._name = database_get_name(database)
+
+    @property
+    def native_object(self):
+        return self._native_object
 
     def name(self) -> str:
+        if not self.native_object.thisown:
```
flyingsilverfin, line 48: What is `.thisown`?

tool/behave_rule.bzl
```diff
@@ -83,9 +83,9 @@ def _rule_implementation(ctx):
              ./typedb_distribution/"$DIRECTORY"/typedb server --server.address 0.0.0.0:$PORT --storage.data typedb_test &
             else
             ./typedb_distribution/"$DIRECTORY"/typedb cluster \
-                --server.address=127.0.0.1:$PORT \
-                --server.internal-address.zeromq=127.0.0.1:$(($PORT+1)) \
-                --server.internal-address.grpc=127.0.0.1:$(($PORT+2)) \
+                --server.address=localhost:$PORT \
```
flyingsilverfin, line 86: I've added an asana task for Dmitrii U to try to upgrade `tonic` so we can use either `localhost` or `127.0.0.1`, since having that option is pretty nice

tests/behaviour/concept/thing/relation/relation_steps.py
```diff
@@ -87,51 +87,52 @@ def step_impl(context: Context, var1: str, role_label: str, var2: str):
...
```
flyingsilverfin, line 95: Ideally the user would only have to catch 1 exception, instead of 2. A "TypeDBClient*" exception is much more intuitive than a sort of native exception "RuntimeError". Let's investigate further if we can get down to 1 error to catch (better UX)

typedb/common/exception.py
```diff
@@ -74,68 +61,98 @@ def __str__(self):
...
     def __init__(self, code: int, message: str):
-        super(ClientErrorMessage, self).__init__(code_prefix="CLI", code_number=code, message_prefix="Client Error", message_body=message)
+        super(ClientErrorMessage, self).__init__(code_prefix="CLI", code_number=code, message_prefix="Client Error",
+                                                 message_body=message)
 
 
-RPC_METHOD_UNAVAILABLE = ClientErrorMessage(1, "The server does not support this method, please check the client-server compatibility:\n'%s'.")
+RPC_METHOD_UNAVAILABLE = ClientErrorMessage(1, "The server does not support this method, please check "
+                                               "the client-server compatibility:\n'%s'.")
 CLIENT_NOT_OPEN = ClientErrorMessage(2, "The client is not open.")
```
flyingsilverfin, line 70: Let's delete unused exceptions.

In general, I'd only expect to see error we throw and wrap from the native layer, and errors we throw to prevent issues in the native layer (eg. passing an illegal value).

typedb/common/streamer.py
```diff
@@ -18,28 +18,20 @@
...
-from abc import ABC, abstractmethod
-from typing import TypeVar, Generic, Callable
 
-T = TypeVar('T')
-U = TypeVar('U')
+from typing import Callable
 
 
-class QueryFuture(Generic[T], ABC):
+class Streamer:
```
flyingsilverfin, line 25: Let's make this generic `Streamer(Generic[T])` returns `T` instances from `__next__()`

flyingsilverfin, line 25: `IteratorWrapper`

flyingsilverfin (changes requested):

BUILD
```diff
@@ -34,14 +34,30 @@ load("@vaticle_dependencies//distribution:deployment.bzl", "deployment")
 load(":deployment.bzl", github_deployment = "deployment")
 
 
+genrule(
+    name = "python-ffi",
```
flyingsilverfin, line 38: typedb-client-ffi

BUILD
```diff
@@ -34,14 +34,30 @@ load("@vaticle_dependencies//distribution:deployment.bzl", "deployment")
...
+    outs = ["typedb/typedb_client_python.py"],
+    srcs = ["@vaticle_typedb_driver_java//rust:typedb_client_python"],
+    cmd = "cp $< $@",
+    visibility = ["//visibility:public"]
+)
+
+genrule(
+    name = "python-lib",
```
flyingsilverfin, line 46: `native_client_binary`

flyingsilverfin: Files and srcs names should be:
`native_client_binary.so`
and
`//rust:_typedb_client_python_binary`

BUILD
```diff
@@ -34,14 +34,30 @@ load("@vaticle_dependencies//distribution:deployment.bzl", "deployment")
 load(":deployment.bzl", github_deployment = "deployment")
 
 
+genrule(
+    name = "python-ffi",
+    outs = ["typedb/typedb_client_python.py"],
```
flyingsilverfin, line 39: Files and srcs should be:
`native_client_wrapper.py`
and
`//rust:typedb_client_python_wrapper`

typedb/concept/value/value.py
```diff
@@ -18,103 +18,114 @@
...
-
-    def get_value_type(self) -> "ValueType":
-        return ValueType.DATETIME
+from typedb.typedb_client_python import value_new_boolean, value_new_long, value_new_double, value_new_string, \
+    value_new_date_time_from_millis, value_is_boolean, value_is_long, value_is_double, value_is_string, \
+    value_is_date_time, value_get_boolean, value_get_long, value_get_double, value_get_string, \
+    value_get_date_time_as_millis
+
+
+class _Value(Value, _Concept):
```
flyingsilverfin, line 36: Discussed in person

flyingsilverfin (changes requested):

typedb/api/concept/type/type.py
```diff
@@ -45,32 +51,25 @@ def is_abstract(self) -> bool:
...
-
     def to_json(self) -> Mapping[str, str]:
         return {"label": self.get_label().scoped_name()}
 
-
-class RemoteType(RemoteConcept, Type, ABC):
-
     @abstractmethod
-    def set_label(self, new_label: str) -> None:
+    def get_supertype(self, transaction: TypeDBTransaction) -> Type:
```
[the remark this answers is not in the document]
dmikhalin: Right, all implementational are `Optional`.

typedb/api/concept/value/value.py
```diff
@@ -18,105 +18,115 @@
...
+    DOUBLE = _ValueType(True, False, Double)
+    STRING = _ValueType(True, True, String)
+    DATETIME = _ValueType(True, True, DateTime)
+
+    @property
+    def native_object(self):
+        return self.value.native_object
+
+    def __str__(self):
+        mapping = {ValueType.OBJECT: "object",
```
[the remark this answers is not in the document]
dmikhalin: We can't use loop here, because we need to specify a string representation for each value. But if it is always a lowercase enum variant, we don't need neither dict, nor loop.

typedb/api/connection/credential.py
```diff
@@ -20,26 +20,24 @@
...
-        self._password = password
-
-        if (tls_root_ca_path is not None and not path.exists(tls_root_ca_path)):
+    def __init__(self, username: str, password: str, *, tls_root_ca_path: Optional[str] = None,
+                 tls_enabled: bool = True):
+        if tls_root_ca_path is not None and not tls_enabled:
+            raise TypeDBClientException.of(CLUSTER_INCONSISTENT_CREDENTIAL)
+        self._native_object = credential_new(username, password, tls_root_ca_path, tls_enabled)
+        if check_error():
+            raise TypeDBClientException("Credential error.")
```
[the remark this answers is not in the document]
dmikhalin: Oh, this check should be removed, we raise an exception if `check_error()`.

typedb/concept/answer/concept_map_group.py
```diff
@@ -19,37 +19,44 @@
...
 
-    def owner(self):
-        return self._owner
+    def concept_maps(self) -> Iterator[ConceptMap]:
+        return map(_ConceptMap, Streamer(concept_map_group_get_concept_maps(self._concept_map_group),
+                                         concept_map_iterator_next))
 
-    def concept_maps(self):
-        return self._concept_maps
+    def __str__(self):
```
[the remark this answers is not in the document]
dmikhalin: `str()` calls `__repr__` if `__str__` is not implemented. Now we everywhere have `__repr__` if default implementation could be unsuitable, and `__str__` is implemented only where we need two different representations.

tests/behaviour/concept/thing/relation/relation_steps.py
dmikhalin: We introduce a class `TypeDBClientException`, generated by SWIG. Exceptions of this class are returned from the rust layer. All exceptions from the python layer are now instances of `TypeDBClientExceptionExt(TypeDBClientException)`, and at the user side we need to check only one exception type.

typedb/common/streamer.py
dmikhalin: We decided in-person that we wouldn't like to have a generic for native types.

BUILD
dmikhalin: Renamed, but not exactly to these names.

BUILD
dmikhalin: Renamed, but not exactly to these names.

typedb/connection/database.py
dmikhalin: `.thisown` indicates who owns the object. If it is set to 1, Python is the owner, otherwise the Rust side is.

typedb/concept/type/thing_type.py
```diff
@@ -18,103 +18,166 @@
...
+            annotations_array,
+        )
+
+    def get_owns(self, transaction: _Transaction, value_type: Optional[ValueType] = None,
+                 transitivity: Transitivity = Transitivity.TRANSITIVE, annotations: Optional[set[Annotation]] = None
+                 ) -> Iterator[AttributeType]:
+        return map(attribute_type_of,
+                   Streamer(thing_type_get_owns(transaction.native_object,
+                                                self.native_object,
+                                                value_type.native_object if value_type else None,
```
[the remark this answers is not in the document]
dmikhalin: In progress

typedb/concept/type/thing_type.py
dmikhalin: In progress

typedb/user/user_manager.py
```diff
@@ -0,0 +1,64 @@
...
+if TYPE_CHECKING:
+    from typedb.api.user.user import User
+    from typedb.native_client_wrapper import Connection as NativeConnection
+
+
+class _UserManager(UserManager):
+
+    def __init__(self, connection: NativeConnection):
+        self._native_user_manager = user_manager_new(connection)
+        self._native_connection = connection
```
flyingsilverfin, line 40: Delete the connection

flyingsilverfin: And get it from the user manager in Rust

flyingsilverfin: Let's put the connection in User in Rust, then delete it from here and in User.

flyingsilverfin (changes requested):

tests/behaviour/concept/thing/relation/relation_steps.py
flyingsilverfin: Amazing

typedb/user/user.py
```diff
@@ -0,0 +1,47 @@
...
+from typedb.api.user.user import User
+from typedb.native_client_wrapper import user_get_username, user_get_password_expiry_seconds, user_password_update
+
+if TYPE_CHECKING:
+    from typedb.native_client_wrapper import User as NativeUser, Connection as NativeConnection
+
+
+class _User(User):
+
+    def __init__(self, user: NativeUser, connection: NativeConnection):
```
flyingsilverfin, line 34: This will hold the native `User` and the Python `UserManager`.

typedb/user/user.py
```diff
@@ -0,0 +1,47 @@
...
+    def username(self) -> str:
+        return user_get_username(self._native_user)
+
+    def password_expiry_seconds(self) -> Optional[int]:
+        if res := user_get_password_expiry_seconds(self._native_user) >= 0:
+            return res
+        return None
+
+    def password_update(self, password_old: str, password_new: str) -> None:
+        user_password_update(self._native_user, self._native_connection, password_old, password_new)
```
flyingsilverfin, line 47: Then we will pass the native `user_manager` instead of `connection`.

Rust side will get the connection from the user manager

typedb/user/user_manager.py
flyingsilverfin, line 40: This will still end up deleted.

typedb/concept/type/thing_type.py
dmikhalin: We should get an exception from the server. Now the error is `The transaction is closed because of the error(s):
status: Internal, message: "", details: [], metadata: MetadataMap { headers: {} }`, and it has to be fixed at the server side.

typedb/concept/type/thing_type.py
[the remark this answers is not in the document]
dmikhalin: Fixed and tested using debug tests. But we should add BDD tests with `value_type != None`.

typedb/api/connection/credential.py
```diff
@@ -20,26 +20,22 @@
...
-
-    def tls_root_ca_path(self):
-        return self._tls_root_ca_path
+    def __init__(self, username: str, password: str, *, tls_root_ca_path: Optional[str] = None,
+                 tls_enabled: bool = True):
+        if tls_root_ca_path is not None and not tls_enabled:
+            raise TypeDBClientExceptionExt.of(CLUSTER_CREDENTIAL_INCONSISTENT)
+        self._native_object = credential_new(username, password, tls_root_ca_path, tls_enabled)
+        if tls_root_ca_path is not None and not path.exists(tls_root_ca_path):
+            raise TypeDBClientExceptionExt.of(CLUSTER_INVALID_ROOT_CA_PATH, tls_root_ca_path)
```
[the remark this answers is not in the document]
dmikhalin: Path validation is unnecessary here.

typedb/concept/answer/concept_map.py
```diff
@@ -19,138 +19,156 @@
...
 
         def __eq__(self, other):
             if other is self:
                 return True
             if not other or type(other) != type(self):
                 return False
-            return self._explainable_id
+            return self.id() == other.id()
 
         def __hash__(self):
```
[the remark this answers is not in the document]
dmikhalin: Fixed (implemented directly because we don't have such native method).

typedb/common/native_object_mixin.py
```diff
@@ -18,28 +18,29 @@
...
-from typing import TypeVar, Generic, Callable
+from typing import Any
 
-T = TypeVar('T')
-U = TypeVar('U')
+from typedb.common.exception import TypeDBClientExceptionExt
 
 
-class QueryFuture(Generic[T], ABC):
+class NativeObjectMixin(ABC):
```
flyingsilverfin, line 30: We don't need the name `mixin`, it's just used as a supertype right?

typedb/concept/answer/concept_map.py
```diff
@@ -19,138 +19,171 @@
...
-class _ConceptMap(ConceptMap):
+class _ConceptMap(ConceptMap, NativeObjectMixin):
 
-    def __init__(self, mapping: Mapping[str, Concept], explainables: ConceptMap.Explainables = None):
-        self._map = mapping
-        self._explainables = explainables
+    def __init__(self, concept_map: NativeConceptMap):
+        if not concept_map:
+            raise TypeDBClientExceptionExt(NULL_NATIVE_OBJECT)
+        self.__native_object = concept_map
```
flyingsilverfin, line 57: If we pass the native handle up to `NativeObjectMixin.__init__(concept_map)`, then we don't need the property in every subtype.

If we want to preserve the return type to be specific, like `NativeConceptMap`, then we use a generic on `NativeObjectMixin[T]`.

typedb/api/connection/credential.py
```diff
@@ -19,27 +19,21 @@
...
-        self._password = password
-
-        if (tls_root_ca_path is not None and not path.exists(tls_root_ca_path)):
-            raise TypeDBClientException.of(CLUSTER_INVALID_ROOT_CA_PATH, tls_root_ca_path)
 
-        self._tls_root_ca_path = tls_root_ca_path
-
-    def username(self):
-        return self._username
+class TypeDBCredential:
```
flyingsilverfin, line 29: This should also be a native object?

typedb/api/connection/options.py
```diff
@@ -18,143 +18,154 @@
...
-        self.read_any_replica = read_any_replica
-        return self
+from typedb.common.exception import TypeDBClientExceptionExt, POSITIVE_VALUE_REQUIRED
 
-    def proto(self) -> options_proto.Options:
-        proto_options = super(TypeDBClusterOptions, self).proto()
 
-        if self.read_any_replica is not None:
-            proto_options.read_any_replica = self.read_any_replica
+class TypeDBOptions:
```
flyingsilverfin, line 40: This should be a `nativeObject`

typedb/api/connection/credential.py
```diff
@@ -19,27 +19,22 @@
...
 
+from typedb.common.exception import TypeDBClientExceptionExt, CLUSTER_CREDENTIAL_INCONSISTENT, ILLEGAL_STATE
+from typedb.common.native_wrapper import NativeWrapper
 
-class TypeDBCredential:
 
-    def __init__(self, username: str, password: str, tls_root_ca_path: str = None):
-        self._username = username
-        self._password = password
+class TypeDBCredential(NativeWrapper[NativeCredential]):
```
flyingsilverfin, line 30: Looks nice!!

flyingsilverfin (approved):

